PR #136, opened 2022-04-15
Support for WolfSSL

maxammann:


src/agent.rs
```diff
@@ -115,6 +115,11 @@ impl From<i32> for TLSVersion {
 /// An [`Agent`] holds a non-cloneable reference to a Stream.
 pub struct Agent {
     pub descriptor: AgentDescriptor,
+
+    #[cfg(feature = "wolfssl")]
+    pub stream: WolfSSL,
+
+    #[cfg(feature = "openssl")]
     pub stream: OpenSSL,
```
maxammann, line 123: I just thought about this problem. Maybe one way to make this cleaner would be to make `Agent` a generic `Agent<S>`, where `S: StreamKind`. The trait `StreamKind` now defines several methods like `fn reset(&self)`.

Then, we could compile-conditional `impl StreamKind for WolfSSL` or `impl StreamKind for OpenSSL`.
That way the `Agent` is decoupled from the stream kind and already generic enough for any stream kind.

maxammann:
The compilation currently fails because of a dependency resolve issue. I think this was not the error you referred to in in mattermost, right?

src/wolfssl_binding.rs
```diff
@@ -72,39 +91,54 @@ impl<S: Read + Write> SslStream<S> {
...
+    unsafe {
+        //// Global WolfSSL lib initialization
+        init();
+
+        //// Context builder
+        wolf::wolfSSL_Init();
+        let ctx: *mut wolf::WOLFSSL_CTX = wolf::wolfSSL_CTX_new(wolf::wolfTLSv1_3_client_method());
+        // Disallow EXPORT in client
+        let cipher_list = CString::new("ALL:!EXPORT:!LOW:!aNULL:!eNULL:!SSLv2").unwrap();
+        wolf::wolfSSL_CTX_set_cipher_list(ctx, cipher_list.as_ptr() as *const _); //
```
LCBH, line 140: Things are simpler than I thought thanks to the OpenSSL compatibility later of WolfSSL that I finally managed to enable (see `build.rs` in the wolfssl-sys crate and the option `--enable-all`).

src/wolfssl_bio.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub stream: S,
+    pub error: Option<io::Error>,
+    pub panic: Option<Box<dyn Any + Send>>,
+    pub dtls_mtu_size: c_long,
+}
+unsafe fn state<'a, S: 'a>(bio: *mut BIO) -> &'a mut StreamState<S> {
+    &mut *(BIO_get_data(bio) as *mut _)
+}
+
+unsafe extern "C" fn bwrite<S: Write>(
```
LCBH, line 114: Again, I heavily rely on the OpenSSL compatibility layer of WolfSSL and I hope that the internal conventions are the same.
Maybe it would have been simpler to link wolfssl to the the rust-openssl crate after all but I failed when I tried this as the latter needs a lot more of functions to be exposed than what we need here.

LCBH (commented):
Currently, the tests for wolfssl unsurprisingly fail given that some methods for WolfSSL PUT in `concretize.rs` are still unimplemented. What is more surprising is that the execution of the (expected) panick panicks itself ;)
Current TODOs:

- [ ] investigate this
- [ ] implement remaining methods in `concretize.rs`
- [ ] tests (happy flow should be executable on WolfSSL PUT)

src/wolfssl_bio.rs
maxammann: Yes, such unsafe code is required sadly.
I think its easier to do it that way. Else we would also have to maintain WolfSSL support in rust-openssl. Or fork it and having troubles to keep it up-to-date.

maxammann: Are we actually using the compatibility layer here? Did you find a C example which uses the wolfssl API instead of the OpenSSL API?

src/wolfssl_binding.rs
maxammann: Good :)

maxammann:
> Currently, the tests for wolfssl unsurprisingly fail given that some methods for WolfSSL PUT in `concretize.rs` are still unimplemented. What is more surprising is that the execution of the (expected) panick panicks itself ;) Current TODOs:
> 
>     * [ ]  (?) investigate this double panick

Yes I was this error often with LibAFL. :D

>     * [ ]  implement remaining methods in `concretize.rs`
> 
>     * [ ]  tests (happy flow should be executable on WolfSSL PUT)

security-claims/Cargo.toml
```diff
@@ -2,10 +2,11 @@
 build = "build.rs"
 name = "security-claims"
 version = "0.1.0"
-
+edition = "2018"
```
maxammann, line 5: Should we upgrade to `2021`

security-claims/Cargo.toml
```diff
@@ -2,10 +2,11 @@
...
 
 [dependencies]
 hex = "0.4"
+itertools = "0.10"
```
maxammann, line 9: Can you list here 1-2 usage of this? So far I was able to survive with `std`

security-claims/src/violation.rs
```diff
@@ -1,11 +1,35 @@
...
-
-use crate::agent::{AgentName, TLSVersion};
-
-type ClaimMessage = (AgentName, Claim);
+use std::any::Any;
+use std::fmt::Debug;
+use std::hash::Hash;
+
+// Will be instantiated with (AgentName,Claim)
+pub type ClaimMessage<AgentName>
```
maxammann, line 8: Why does `AgentName` need to be generic?

security-claims/src/violation.rs
```diff
@@ -1,11 +1,35 @@
...
+where
+    AgentName: Copy + Clone + Debug + Eq + PartialEq + Hash,
```
maxammann, line 10:
```suggestion
    A: AgentName,
```

We should create here a trait `AgentName`:

```rust
trait AgentName: Copy + Clone + Debug + Eq + PartialEq + Hash {}
```

security-claims/src/violation.rs
```diff
@@ -1,11 +1,35 @@
...
+= (AgentName, Claim);
+
+// This code is duplicated from super::super::agent, which is unfortunate.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum TLSVersion {
+    V1_3,
+    V1_2,
+    Unknown,
+}
+// This code is duplicated from super::super::agent, which is unfortunate.
```
maxammann, line 20: Can we reuse `TLSVersion` in `agent`?

src/agent.rs
```diff
@@ -103,35 +114,38 @@ pub enum TLSVersion {
     Unknown,
 }
 
-impl From<i32> for TLSVersion  {
+impl From<i32> for TLSVersion {
```
maxammann, line 117: I think we should reuse `TLSVersion` from `security-claims` crate.

maxammann: Or create a `types` create.

security-claims/src/violation.rs
```diff
@@ -1,11 +1,35 @@
+use crate::{Claim, ClaimCipher, ClaimType};
```
maxammann, line 1: I'm not sure if its good that `violation.rs` is part of `security-claims`. The violation check is part of the fuzzer I think. The `security-claims` package only exists because of the header file and the custom `build.rs` script which I did not want to include in the main fuzzer.

src/agent.rs
```diff
@@ -103,35 +114,38 @@ pub enum TLSVersion {
...
 /// An [`Agent`] holds a non-cloneable reference to a Stream.
 pub struct Agent {
     pub descriptor: AgentDescriptor,
-    pub stream: OpenSSLStream,
+    pub stream: Box<dyn PUT>,
 }
 
 impl Agent {
-    pub fn new_openssl(
+    pub fn new<P: 'static + PUT>(
```
maxammann, line 134: Minor:
```suggestion
    pub fn new<P: PUT>(
```

I think its simpler to add the `'static` to the `PUT` trait.

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+use foreign_types_shared::ForeignTypeRef;
+use rustls::msgs::message::OpaqueMessage;
+use security_claims::{deregister_claimer, register_claimer, Claim};
+use serde::{Deserialize, Serialize};
+use std::cell::RefCell;
+use std::io::{Read, Write};
+use std::rc::Rc;
+
+/// Stream, Read, Write traits below are with respect to this content type // [TODO:PUT] how one can make this precise in the type (Without modifing those traits specs?)
+pub type Bts<'a> = &'a [u8];
```
maxammann, line 23: For what stands the abbreviation `Bts`?

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::{io, openssl_binding};
+use foreign_types_shared::ForeignTypeRef;
+use rustls::msgs::message::OpaqueMessage;
+use security_claims::{deregister_claimer, register_claimer, Claim};
+use serde::{Deserialize, Serialize};
+use std::cell::RefCell;
+use std::io::{Read, Write};
+use std::rc::Rc;
+
+/// Stream, Read, Write traits below are with respect to this content type // [TODO:PUT] how one can make this precise in the type (Without modifing those traits specs?)
```
maxammann, line 22: I don't understand the TODO here, what do you mean?

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+    ///
+    pub claimer: Rc<RefCell<VecClaimer>>,
+}
+
+#[derive(Debug, Copy, Clone, Deserialize, Serialize, Eq, PartialEq)]
+pub enum PUTType {
+    OpenSSL,
+    WolfSSL,
+}
+pub trait PUT: Stream + Drop {
```
maxammann, line 41: Looks good!

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Returns a textual representation of the state in which self is
+    fn describe_state(&self) -> &'static str;
+    /// Returns a textual representation of the version of the PUT used by self
+    fn version(&self) -> &'static str;
+    /// Make the PUT used by self determimistic in the future by making its PRNG "deterministic"
+    fn make_deterministic(&self) -> ();
+}
+
+pub fn put_version(put_type: PUTType) -> &'static str {
+    let c = Config {
```
maxammann, line 65:
```suggestion
    let config = Config {
```

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+        tls_version: TLSVersion::V1_3,
+        server: false,
+        agent_name: AgentName::new(),
+        claimer: Rc::new(RefCell::new(VecClaimer::new())),
+    };
+    let put: Box<dyn PUT> = match { put_type } {
+        OpenSSL => Box::new(OpenSSL::new(c).expect("Failed to create a put instance")),
+        WolfSSL => Box::new(WolfSSL::new(c).expect("Failed to create a put instance")),
+    };
+    put.as_ref().version()
```
maxammann, line 75: I think its fine here to construct a whole PUT because `put_version` is only used for debugging.

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+        tls_version: TLSVersion::V1_3,
+        server: false,
+        agent_name: AgentName::new(),
+        claimer: Rc::new(RefCell::new(VecClaimer::new())),
+    };
+    let put: Box<dyn PUT> = match { put_type } {
+        OpenSSL => Box::new(OpenSSL::new(c).expect("Failed to create a put instance")),
+        WolfSSL => Box::new(WolfSSL::new(c).expect("Failed to create a put instance")),
+    };
+    put.as_ref().make_deterministic()
```
maxammann, line 89: As this function is called during fuzzing this might be a litte bit slow.

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+//! - [`progress`] makes a state progress (interacting with the buffers)
+//!
+//! And specific implementations of PUT for the different PUTs.
+use crate::agent::{AgentName, TLSVersion};
+use crate::error::Error;
+use crate::io::MessageResult;
+use crate::io::{MemoryStream, Stream};
+use crate::openssl_binding::openssl_version;
+use crate::trace::VecClaimer;
+use crate::wolfssl_binding;
```
maxammann, line 12: Can be done later: I think we should make it possible to enable/disable openssl or wolfssl via feature flags.

* Compiling wolfssl & openssl takes a long time to compile
* By depending expicitely on wolfssl and openssl we make a "closed set" of supported PUTs. At some point we probably will have to make `tlspuffin` a library instead of a binary. Then library consumers can add support for any protocol to `tlspuffin`. That will probably be a task for my internship.

src/concretize.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+////////////// WolfSSL specific-state
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub struct WolfSSL {
+    stream: wolfssl_binding::SslStream<MemoryStream>,
+}
```
maxammann, line 199: This should probably go into a separate file.

src/fuzzer/libafl_setup.rs
```diff
@@ -1,7 +1,6 @@
 use core::time::Duration;
```
maxammann, line 1: I think except formatting nothing changed here.

src/fuzzer/seeds.rs
```diff
@@ -2,23 +2,25 @@
...
     trace::{Action, InputAction, OutputAction, Step, Trace},
 };
-use rustls::{ProtocolVersion, CipherSuite};
-use rustls::msgs::handshake::{SessionID, Random};
 use rustls::internal::msgs::handshake::ServerExtension;
+use rustls::msgs::handshake::{Random, SessionID};
+use rustls::{CipherSuite, ProtocolVersion};
 
-pub fn seed_successful(client: AgentName, server: AgentName) -> Trace {
+pub fn seed_successful(client: AgentName, server: AgentName, put_type: PUTType) -> Trace {
```
maxammann, line 23: Just a quick idea: We could make `AgentName` generic over a PUT. That way we could infer which PUT to spawn based on the AgentName.

maxammann: The serializability of traces would be a problem then though.

src/trace.rs
```diff
@@ -370,27 +373,37 @@ impl TraceContext {
     }
 
     pub fn new_openssl_agent(&mut self, descriptor: &AgentDescriptor) -> Result<AgentName, Error> {
-        let agent_name = self.add_agent(Agent::new_openssl(descriptor, self.claimer.clone())?);
+        // [TODO] There might be a cleaner way of doing this, e.g., if we could store a type
+        // instantiation in an enum, but this works
+        let agent = match descriptor.put_type {
+            PUTType::OpenSSL => Agent::new::<OpenSSL>(descriptor, self.claimer.clone())?,
+            PUTType::WolfSSL => Agent::new::<WolfSSL>(descriptor, self.claimer.clone())?,
+        };
```
maxammann, line 381: I think we need at some point a cleaner method, else we can not disable WolfSSL/OpenSSL via a feature flag.

src/trace.rs
```diff
@@ -469,6 +482,7 @@ impl Trace {
 
         let claims: &Vec<(AgentName, Claim)> = &ctx.claimer.deref().borrow().claims;
         if let Some(msg) = is_violation(claims) {
+            // [TODO] versus checking at each step ? Could detect violation earlier, before a blocking state is reached ? [BENCH] benchmark the efficiency loss of doing so
```
maxammann, line 485: Not sure if that can be done. But actually the idea way to check for violation at each step, because some violations maybe are only visible in-between a flight of messages.

src/wolfssl_binding.rs
```diff
@@ -0,0 +1,180 @@
+use std::any::Any;
```
maxammann, line 1: Not reviewed :D

src/wolfssl_bio.rs
```diff
@@ -0,0 +1,252 @@
+use libc::{c_char, c_int, c_long, c_void, strlen};
```
maxammann, line 1: Not reviewed :D

maxammann (commented):
I reviewed this to get a better overview. I think all in all this looks good and we also also address all of the comments in a separate PR. I think most stuff would be easy to change, but requires some architectural thinking first.

I think it would actually make sense to generalize `tlspufifn` now to work for arbitrary PUTs and protocols.

Notes:
* Instead of updating `bindgen` in `deps/rust-openssl-src` I think we should be able to simply override the dependency version: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

src/trace.rs
LCBH: Maybe I did not understand how those checks are performed. It seemed to me that there were done once a trace is fully executed (?), in which case we might miss attacks, e.g., if an agent thinks that he has authenticated a peer, while he has not properly yet authenticated it.

maxammann: Oh sorry, yes, the checks are executed after the trace is done. But on the data which has been recorded during the trace execution. So no attacks should be missed.

src/concretize.rs
```diff
@@ -200,33 +200,35 @@ pub struct WolfSSL {
 
 impl Stream for WolfSSL {
     fn add_to_inbound(&mut self, result: &OpaqueMessage) {
-       todo!() // self.stream.get_mut().add_to_inbound(result)
+        let a = self.stream.get_mut();
+        a.add_to_inbound(result)  // SEGFAULT: here a = NULL when we execute the first input, which triggers add_to_inbound here! Somehow, bio initialization fails !
```
LCBH, line 204: Segfault here when running `cago test --package tlspuffin --lib tests::seeds::test_seed_client_attacker12 --no-default-features --features wolfssl520 -- --exact`

maxammann: Feel free to tell me if you need support here :)
I'm quite busy with another project right now, but can always support.

security-claims/src/violation.rs
LCBH: To void a circular dependency (agent::AgentName is in agent, which depends on security-claims).

src/fuzzer/seeds.rs
LCBH: agent::AgentDescriptor does that, it now has a field `pub put_type: PUTType`. PUTType is just an enum so no problem for serializability. It just keeps track of which PUT will have to be used for executing/animating that agent.

src/trace.rs
LCBH: Hmmm, that's a good point. I wasn't in my list of requirements but it should have been :(

LCBH:
Current status is:

1. this PR is blocked by segfaults regarding BIO initialization. See https://github.com/tlspuffin/tlspuffin/pull/136/commits/65349e6224662fa574ee4675ce7c5223751af1c4   Debugging this reveals 
2. Since I thought it would be hard to debug this because this PR is a bunch of unsafe code, I created this other PR: https://github.com/tlspuffin/tlspuffin/pull/139 where I tried to adapt rust-openssl for wolfssl. Very surprisingly, I encountered the same SEGFAULT that I am in the process of debugging.

PR #139, opened 2022-05-13
Support for WolfSSL through a new rust-wolfssl crate

LCBH:
I tried another approach as I struggled with #136 to debug some SEGFAULT.

Here the approach is to clone rust-openssl, minimize the exposed interface while exposing what we need for `wolfssl_binding.rs`, then plug in wolfssl-sys instead of openssl-sys.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,3 +14,7 @@
 	path = deps/wolfssl-sys
 	url = git@github.com:tlspuffin/wolfssl-sys.git
 	branch = openssl
+[submodule "deps/rust-wolfssl"]
+	path = deps/rust-wolfssl
+	url = git@github.com:tlspuffin/rust-openssl.git
+	branch = wolfssl
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1804,7 +1804,7 @@ dependencies = [
  "test-log",
  "webpki",
  "webpki-roots",
- "wolfssl-sys",
+ "wolfssl",
 ]
 
 [[package]]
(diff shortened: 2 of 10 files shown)


LCBH (commented):
Super weird experiment:
`cargo test --package tlspuffin --lib tests::seeds::test_seed_client_attacker12 --no-default-features --features openssl111 -- --exact`
works just fine while
`cargo test --package tlspuffin --lib tests::seeds::test_seed_client_attacker12 --no-default-features --features wolfssl520 -- --exact`
triggers the same NULL pointer access I got with the other approach from #136 .
This is surprising since the change of features change the crate used in {wolfssl,openssl}_binding.rs: either rust-openssl or rust-wolfssl, which should be semantically identical!

maxammann:
This approach will not work: The wolfssl bindings are just `#defines` from for example `SSL_new_ctx` to `wolfSSL_new_ctx`. The linker can not use those "translatoins". Therefore your previous approach is the way to go: We need to use the `wolfSSL*` functions directly.

maxammann:
Lets discuss this in Mattermost. So the main reason is that the compatibility is only source-level. LibreSSL is ABI-compatible with OpenSSL.

The linker is not able to link to `wolfSSL_CTX_new` when writing `SSL_CTX_new` because the linker can not translate between the two. It would be possible to do this by providing a mapping file to the linker.
But at that point it gets very hairy and also the source-compatability is not 100%, but only 99%.

PR #147, opened 2022-06-20
Trace-dependent security policies [bug oracle without claims]

LCBH:
As explained in #148, the aim is to develop a complementary bug oracle that solely depends on the trace and the exchanged messages.

### DIFF
--- a/tlspuffin-claims/src/violation.rs
+++ b/tlspuffin-claims/src/violation.rs
@@ -108,10 +108,10 @@ where
                     }
 
                     // TODO@MAX: Do you agree that this policy below will capture server authentication bypass attacks?
-                    if client.peer_cert.pkey != server.cert.pkey {
+                    if client.peer_cert.data != server.cert.data {
                         trace!("Client peer cert: {:?}", client.peer_cert);
                         trace!("Server cert {:?}", server.cert);
-                        return Some("Mismatching server certificate");
+                        return Some("Mismatching server certificate hash");
                     }
 
 
--- a/tlspuffin-claims/src/violation.rs
+++ b/tlspuffin-claims/src/violation.rs
@@ -1,6 +1,7 @@
 use std::{any::Any, fmt::Debug, hash::Hash};
 
 use itertools::Itertools;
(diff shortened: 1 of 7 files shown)


tlspuffin-claims/src/violation.rs
```diff
@@ -106,6 +107,14 @@ where
                         return Some("Mismatching ciphers");
                     }
 
+                    // TODO@MAX: Do you agree that this policy below will capture server authentication bypass attacks?
```
maxammann, line 110: I agree that if a trace succeeds (both puts are in a finished state) and those two do not match, then we have a MITM.

I'm not 100% sure whether an attack can be observed this was though. E.g. its unclear whether the `pkey` will always be initialized or not given the implementation which writes to it.
This would require a test case to validate this I think.

Maybe we could use a known wolfssl bug to verify this.

maxammann (commented):
btw: I found a waaay better method to instrument the PUTs without direcly modifying the source code of OpenSSL/LibreSSL.

Instead of modifying the source code its usually possible to generate Rust bindings from an `internal.h` or in case of OpenSSL `ssl_local.h`: https://github.com/openssl/openssl/blob/cac250755efd0c40cc6127a0e4baceb8d226c7e3/ssl/ssl_local.h#L1212

Then we can directly access the state from Rust. The msg callback allows us to extract data inbetween messages: https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_msg_callback.html

maxammann:
Exactly, using vscode you even get autocomplete. Modifying the source code directly just causes a mess git submodules.

maxammann:
Its still put dependant, but abstractions could be built in rust which simplify the instrumentation

LCBH:
> Its still put dependant, but abstractions could be built in rust which simplify the instrumentation

OK I like it.

I can now use this to check matching conversation (based on the hash transcript, so actually checking agreement on the hash transcript). I'll do that today.

maxammann:
Closing because it is out of date.

PR #163, opened 2022-08-23
Discussion about Lucca's commits before going upstream

LCBH:


puffin/src/agent.rs
```diff
@@ -141,7 +141,7 @@ pub enum TLSVersion {
 pub struct Agent<PB: ProtocolBehavior> {
     name: AgentName,
 
-    put: Box<dyn Put<PB>>,
+    pub put: Box<dyn Put<PB>>,
```
LCBH, line 144: had to do that in order to access shutdown statut of the agent's PUT

tlspuffin/src/tcp/mod.rs
```diff
@@ -794,4 +808,42 @@ mod tests {
 
         info!("{}", shutdown);
     }
+
+    #[test]
+    fn test_wolfssl_buffer_under_read() {
```
LCBH, line 813: This test currently fail with:
```
thread 'tcp::tests::test_wolfssl_buffer_under_read' panicked at 'called `Result::unwrap()` on an `Err` value: DeserializeBadVarint', tlspuffin/src/tcp/mod.rs:832:71
```
but the deseralized trace has been produced on the same machine with the same build options :(

LCBH (commented):
Problem with deserializing a trace

tlspuffin/src/tcp/mod.rs
LCBH: Commit #204086e seems to fix it, but why ? Anyway

LCBH (commented):
Trace has been minized to one hello retry request and still triggers the buffer over read:
```
RUST_BACKTRACE=1 RUSTFLAGS=-Awarnings  RUST_LOG=info RUSTFLAGS=-Zsanitizer=address cargo +nightly test --target x86_64-unknown-linux-gnu --bin tlspuffin -p tlspuffin --features wolfssl540,asan --lib tcp::tests::test_wolfssl_buffer_under_read -- --nocapture
```

PR #164, opened 2022-09-01
Add constant function symbols

maxammann:
To discuss: this is maybe not needed but I thought it could be non attacking-preserving to restrict the adversary to use a single session random.

Moved from: https://github.com/trailofbits/tlspuffin/pull/30


maxammann:
Closing because this is out of date, and also we did not use this in the paper evaluaiton. @LCBH

PR #165, opened 2022-09-01
Use smallvec in hot areas

maxammann:


maxammann:
Unneeded optimization

PR #166, opened 2022-09-01
Tracing

maxammann:


maxammann:
Closing because we do not need tracing support right now. The fuzzer works as expected right now.

The JSON intrinics support is fine.

PR #169, opened 2022-09-01
Fix openssl sys

maxammann:
This is maybe invalid, close if it has no changes


maxammann:
Cherry picked in https://github.com/tlspuffin/tlspuffin/pull/201

PR #215, opened 2022-10-28
DO NOT MERGE:  Reproduce SIG in July Branch

maxammann:
Easily finds SIG


maxammann:
SIG is discoverable in this branch as well as the evaluate branch.

PR #217, opened 2022-10-31
Evaluation

maxammann:
This evaluation is for the first version of the tlspuffin paper.

Execute it using ./evaluate.sh


maxammann:
TODO:
* Revert "Use in-memory corpus"
* Revert "Check security violations after each step"

PR #225, opened 2022-11-03
DO NOT MERGE: Repro sig aflmap

maxammann:


maxammann:
It is unclear whether AflMapFeedback is better than MaxMinFeedback. CLosing as using AFlMapFeedback was only a test.

Issue #204, opened 2022-10-12
OpenSSL 1.0.1 is not building if  OpenSSL 3.0.0 is installed


maxammann:
```
  makedepend: warning:  md5test.c (reading ../include/openssl/md5.h, line 63): cannot find include file "stddef.h"
  	not in ../stddef.h
  	not in ../include/stddef.h
  	not in /usr/include/stddef.h
  makedepend: warning:  hmactest.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
  makedepend: warning:  wp_test.c (reading ../include/openssl/whrlpool.h, line 5): cannot find include file "stddef.h"
  	not in ../stddef.h
  	not in ../include/stddef.h
  	not in /usr/include/stddef.h
  makedepend: warning:  rc2test.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
  makedepend: warning:  rc4test.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
  makedepend: warning:  rc4test.c (reading ../include/openssl/rc4.h, line 67): cannot find include file "stddef.h"
  	not in ../stddef.h
  	not in ../include/stddef.h
  	not in /usr/include/stddef.h
  makedepend: warning:  rc5test.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
  makedepend: warning:  shatest.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
  makedepend: warning:  sha1test.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
  makedepend: warning:  mdc2test.c (reading ../e_os.h, line 449): cannot find include file "sys/types.h"
  	not in ../sys/types.h
  	not in ../include/sys/types.h
  	not in /usr/include/sys/types.h
[...]

PR #234, opened 2023-04-24
Update CI runner

maxammann:
Should maybe wait until pesto is updated

fails because of https://github.com/tlspuffin/tlspuffin/issues/204

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -14,6 +14,22 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: setup CI env
+      shell: bash
+      run: |
+        #!/usr/bin/env bash
+        export_env() {
+          local var=${1}
+          local val=${2:-${!1}}
+
+          printf '%s=%s\n' "${var}" "${val}" >> "${GITHUB_ENV}"
+        }
+
+        export_env RUST_LOG "info"
+        export_env CLANG_PATH "$(clang -print-prog-name=clang)"
+        export_env LIBCLANG_PATH "$(llvm-config-"$(clang -dumpversion | cut -d '.' -f1)" --libdir)"
+        export_env BINDGEN_EXTRA_CLANG_ARGS "-v"
+
     - uses: extractions/setup-just@v1
     - name: Install toolchain
       shell: bash
--- a/.github/workflows/build-deploy-docs.yml
+++ b/.github/workflows/build-deploy-docs.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   build-docs:
     name: Build
-    runs-on: ubuntu-20.04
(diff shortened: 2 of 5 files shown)


LCBH:
> Should maybe wait until pesto is updated
> 
> fails because of #204

What kind of update do you expect on pesto? (I can do it!)

maxammann:
> > Should maybe wait until pesto is updated
> > fails because of #204
> 
> What kind of update do you expect on pesto? (I can do it if it is not 20->22!)

We should not update pesto from 20->22 :) But this PR can only merged after that.

michaelmera:
The linux distro on that particular server changed since this PR was opened and I failed to reproduce #204.

@maxammann Is there anything else preventing this PR to land in main?

maxammann:
If tests pass then we can go ahead. But maybe we should wait until the pesto server updates.

It is nice to have binary comparability between those.

In fact I think we really should wait for pesto...

cc @LCBH

michaelmera:
> If tests pass then we can go ahead. But maybe we should wait until the pesto server updates.
> 
> It is nice to have binary comparability between those.
> 
> In fact I think we really should wait for pesto...
> 

@maxammann I asked around and the pesto server update has already been done some months ago. However, it's not running Ubuntu anymore but Debian, so I wouldn't rely on it for binary comparability.

I will rebase this PR on main and merge it if there are no other objection.

maxammann:
> > If tests pass then we can go ahead. But maybe we should wait until the pesto server updates.
> > It is nice to have binary comparability between those.
> > In fact I think we really should wait for pesto...
> 
> @maxammann I asked around and the pesto server update has already been done some months ago. However, it's not running Ubuntu anymore but Debian, so I wouldn't rely on it for binary comparability.
> 
> I will rebase this PR on main and merge it if there are no other objection.

In this case it is fine! I think there is no way to use GH-hosted runners with debian.

Feel free to merge.

michaelmera:
> But of course things are never that easy. CI checks fail, but only for wolfssl430. I will have a look... 😢

So it appears that `bindgen` is using `clang-sys` which always [prefer the highest version of libclang](https://github.com/KyleMayes/clang-sys/tree/v1.6.1?tab=readme-ov-file#versioned-dependencies) available on the system.

The github-hosted runner ubuntu-22.04 has libclang-13, libclang-14 and libclang-15 installed but the default clang version is 14. So bindgen tries to use libclang-15 with clang-14 and those are compatible enough that everything is building except wolfssl430.

I added a fix to force the consistent pair clang-14/libclang-14 in the CI setup action.

maxammann:
Sounds good! Clang 14 is the one to go for now.

maxammann:
Thanks for those insights!

maxammann:
Definitely a new reason to switch to Nix in the future :)

Issue #279, opened 2023-09-07
Bit-level mutations through lazy bitstring replacements


LCBH:
One approach to integrate bit-level mutations (like HAVOC) to tlspuffin and DY fuzzing in general is as follows:
1. Have a mutation `make_bitstring` that:
  a. randomly choose a sub-term `t` in a recipe in the current trace (similar to what is done in [this clone](https://github.com/nimics/tlspuffin/blob/main/puffin/src/fuzzer/mutations.rs#L598), 
  b. evaluate it to a bitstring `b`,
  c. associate a mutable version of `b`, amenable to bit-level mutations, to `t` in fields `payload` and `initial_payload` of `t`. From now on, the idea is that future evaluation of `t` will use `payload` instead of using the evaluation of `t` and bit-level mutations can mutate in-place the field `payload`.
2. Have bit-level mutations like `HAVOC` operating on all `payloads` of all sub-terms of all recipes in the trace.
3. When evaluating the trace and one of its recipe `t`, do the following to 'b_t':
  a. evaluate it to `b_t` as currently done.
  b. for each sub-term `t'` of `t` having a payload `b` (hence `make_message` has been used on `t'`, do the following:
     - replace `t.initial_payload` that must be found in `b_t` by `t.payload` (even if the two do not have the same size)
     - [Optional] if multiple `t.initial_payload` can be found at different locations, find the right location corresponding to `t'`. (One option is to evaluate the first child of `t` that has `t'` as descendant, evaluate this child as a bitstring, and find `b_t` in it. If there is only one matching location, then we win, otherwise, we recursively do the same on this child (instead of `t`).
4. send the modified `b_t` to the suitable agent.

LCBH:
WIP #348

PR #280, opened 2023-09-12
Term with payloads

LCBH:
See [#279 ]
One approach to integrate bit-level mutations (like HAVOC) to tlspuffin and DY fuzzing in general is as follows:
1. Have a mutation `make_bitstring` that:
  a. randomly choose a sub-term `t` in a recipe in the current trace (similar to what is done in [this clone](https://github.com/nimics/tlspuffin/blob/main/puffin/src/fuzzer/mutations.rs#L598), 
  b. evaluate it to a bitstring `b`,
  c. associate a mutable version of `b`, amenable to bit-level mutations, to `t` in fields `payload` and `initial_payload` of `t`.
  The idea is that, from now on, future evaluation of `t` will use `payload` instead of using the evaluation of `t` and bit-level mutations can mutate in-place the field `payload`.
2. Have bit-level mutations like `HAVOC` operating on all `payloads` of all sub-terms of all recipes in the trace.
3. When evaluating the trace and one of its recipe `t`:
  a. evaluate it to `b_t` as currently done.
  b. for each sub-term `t'` of `t` having a payload `b` (hence `make_message` has been used on `t'`), do the following:
     - replace `t.initial_payload` that must be found in `b_t` by `t.payload` (even if the two do not have the same size)
     - [Optional] if multiple `t.initial_payload` can be found at different locations, find the right location corresponding to `t'`. (One option is to evaluate the first child of `t` that has `t'` as descendant, evaluate this child as a bitstring, and find `b_t` in it. If there is only one matching location, then we win, otherwise, we recursively do the same on this child (instead of `t`).
5. Send the modified `b_t` to the suitable agent. 

Pros:
 - compared to [#278], this approach allows to accept any kind of bit-level mutations. Approaches based on re-interpreting the mutated bit-strings in the Mapper internal structure (here `rustls`) rejects a lot of mutations that make this re-interpretation/parsing fail. Problem 1: We are not really interested in fuzzing the Mapper: a mutation that may be gracefully rejected by the Mapper might crash the PUT's "Mapper"/parsing routines. Problem 2: even if we used the PUT as Mapper, the context of evaluation in our fuzzer will not be the same as the one of the agent we fuzz, hence some bugs may be missed because of we reject a mutation on the fuzzer side (parsing fails) but this would still crash the PUT.
 --> This approach does not suffer from those problems, any bit-level mutation will go through.
 - we are able to fuzz any sub-terms with this approach. On the contrary, approaches based on fuzzing `rustls payloads`  (like [this clone](https://github.com/nimics/tlspuffin/blob/main/puffin/)) are only able to fuzz certain parts of messages (such as certificates). Problem 3: structures of messages (header, length, etc.) cannot be fuzzed this way.
 
 Cons:
  - quite cumbersome and hacky
  - tracking down which precise location of the bitstring `b` to replace is non-trivial (but possible)
  - we need to re-architecture the interface tlspuffin-puffin since no bitstring-access to the PUT agents are exposed (for sending bitstring), only message-level-access to the PUT is exposed
  - maybe too inefficient once we have a lot of sub-terms with payloads (?)
  
Question: some HAVOC sub-mutations are more expressive if we give them the concatenation of all `payloads` of all sub-terms (e.g., the ones that swap around some parts of the bitstrings). We may want do to that but reconstructing the new mutated `payloads`  after having applied the mutation is non-trivial.

### DIFF
--- a/puffin/src/fuzzer/mutations.rs
+++ b/puffin/src/fuzzer/mutations.rs
@@ -6,6 +6,7 @@ use crate::{
     fuzzer::term_zoo::TermZoo,
     trace::Trace,
 };
+use crate::algebra::{TermEval, TermType};
 
 pub fn trace_mutations<S, M: Matcher>(
     min_trace_length: usize,
@@ -74,7 +75,8 @@ where
         if let Some((term_a, trace_path_a)) = choose(trace, self.constraints, rand) {
             if let Some(trace_path_b) = choose_term_path_filtered(
                 trace,
-                |term: &Term<M>| term.get_type_shape() == term_a.get_type_shape(),
+                |term: &TermEval<M>| term.get_type_shape() == term_a.get_type_shape(),
+                // TODO: maybe also check that both terms are .is_symbolic()
                 self.constraints,
                 rand,
             ) {
@@ -136,10 +138,12 @@ where
         _stage_idx: i32,
     ) -> Result<MutationResult, Error> {
         let rand = state.rand_mut();
-        let filter = |term: &Term<M>| match term {
+        let filter = |term: &TermEval<M>| match &term.term {
             Term::Variable(_) => false,
-            Term::Application(_, subterms) => subterms
-                .find_subterm(|subterm| match subterm {
+            Term::Application(_, subterms) =>
+                // TODO: maybe add: term.is_symbolic() &&
+                subterms
+                .find_subterm(|subterm| match &subterm.term {
                     Term::Variable(_) => false,
                     Term::Application(_, grand_subterms) => {
                         grand_subterms.find_subterm_same_shape(subterm).is_some()
@@ -149,7 +153,8 @@ where
         };
         if let Some(mut to_mutate) = choose_term_filtered_mut(trace, filter, self.constraints, rand)
         {
-            match &mut to_mutate {
+            match &mut to_mutate.term {
+                // TODO: maybe also SKIP if not(to_mutate.is_symbolic())
                 Term::Variable(_) => Ok(MutationResult::Skipped),
                 Term::Application(_, ref mut subterms) => {
                     if let Some(((subterm_index, _), grand_subterm)) = choose_iter(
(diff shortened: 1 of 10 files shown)


maxammann:
I changed the Github PR base to the BoringSSL branch to clean things up.

btw if you want to link your commits to your GH account you need to add lucca.hirschi@inria.fr to your GH account ;)

maxammann:
Will review when all other PRs are merged.

LCBH:
> Will review when all other PRs are merged.

This still WIP though, but I'd be happy to have any feedback.

puffin/src/fuzzer/mutations.rs
```diff
@@ -602,9 +602,12 @@ where
     // Only execute shorter trace: trace[0..step_index])
     // execute the PUT on the first step_index steps and store the resulting trace context
     tr.execute_until_step(ctx, path.0).err().map(|e| {
-        error!("mutation::MakeMessage trace is not executable until step {},\
+        // 20% to 50% MakeMessage mutations fail, so this is a bit costly :(
+        // TODO: we could memoize the recipe evaluation in a Option<ConcreteMessage> and use that
```
LCBH, line 606: This is worth being addressed in the future

LCBH:
https://github.com/tlspuffin/tlspuffin/pull/280/commits/52a6d3b6c9e6f71ec402c61ecce92708490bcafa solves most inconsistencies when looking for payloads. No more `Payloads returned by eval_until_opaque were inconsistent!` after 6785 executions.

puffin/src/algebra/dynamic_function.rs
```diff
@@ -67,6 +67,32 @@ use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
...
+    /// conretization of its arguments
+    pub is_opaque: bool,
+    /// Whether the function symbol computes a list such as `fn_append_certificate`.
+    pub is_list: bool,
+    /// Whether the function symbol computes a strict sub-term (accessed function symbols).
+    /// Incidentally, its concretization does not contain all the conretizations of its arguments.
+    /// Examples: `fn_get_server_key_share`.
+    pub is_get: bool,
+}
+// TODO: add a uni test for making sure the given attributes are correct
```
LCBH, line 84: TODO: add a uni test for making sure the given attributes are correct (`pub struct FunctionAttributes`)

PR #283, opened 2023-10-18
Add support for OpenSSL 3

maxammann:


LCBH:
@maxammann Do you want to update and merge?

maxammann:
Yes, I'll review this again and then merge! :)

maxammann:
Hm, not sure why these builds are failing

michaelmera (commented):
I think this is the change that breaks part of the CI: before OpenSSL 1.1.1 the libdir argument is always relative to the prefix.

What was the reason for adding this change?

see also upstream:
- issue: https://github.com/openssl/openssl/issues/5398
- fixed by: https://github.com/openssl/openssl/commit/e454f3add638fda5c2aa32cd368c8929c0b1eb09

PR #284, opened 2023-10-20
Add TCP command

maxammann:


LCBH:
We should rebase this onto `main` in a new PR.

PR #290, opened 2023-11-24
Support for boringssl

aeyno:
Adding support for BoringSSL PUT


Cargo.toml
```diff
@@ -48,3 +51,4 @@ debug = true
 
 [profile.dev]
 panic = "abort"
+lto = true
```
michaelmera, line 54: Is there a reason for switching link time optimizations on?

michaelmera (changes requested):
I did a quick review and the PR seems fine. I have not spent a lot of time to analyze the PUT implementation itself, but it seemed mostly similar to what we had in OpenSSL/WolfSSL, except for the way messages are handled.

In my opinion, the changes regarding determinism/reseed would have been better in a different PR, but it seems to work fine for now and we will need to review that anyway when implementing PUT "capabilities" in the future.

See my inline comments for minor changes before merging. Additionally, you will need to change `tlspuffin/build.rs` to prevent the selection of `boringssl-binding` together with either `wolfssl-binding` or `openssl-binding`.

Cargo.toml
aeyno: Yes, we had a problem where we were only able to build in release and not debug (ld errors) and we found out that turning on lto on debug solved the problem. I don't remember what the exact problem was but there is maybe another solution

tlspuffin/Cargo.toml
```diff
@@ -187,16 +187,56 @@ wolfsslmaster = [
...
-sancov_libafl = ["puffin/sancov_libafl", "openssl-src?/sancov", "wolfssl-sys?/sancov"]
+sancov_libafl = ["puffin/sancov_libafl", "openssl-src?/sancov", "wolfssl-sys?/sancov", "boringssl-sys?/sancov"]
 
 # Enables ASAN
-asan = ["openssl-src?/asan", "wolfssl-sys?/asan"]
+asan = ["openssl-src?/asan", "wolfssl-sys?/asan", "boringssl-sys?/asan"]
 
-gcov_analysis = ["openssl-src?/gcov_analysis", "wolfssl-sys?/gcov_analysis"]
-llvm_cov_analysis = ["openssl-src?/llvm_cov_analysis", "wolfssl-sys?/llvm_cov_analysis"]
+gcov_analysis = ["openssl-src?/gcov_analysis", "wolfssl-sys?/gcov_analysis", "wolfssl-sys?/gcov_analysis"]
```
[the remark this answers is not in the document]
aeyno: Yes, will change that

Cargo.toml
michaelmera: Sorry for reopening that, but I just tried and I am able to build in debug mode without lto.

The additional link time introduced is not insignificant (with 64 cores/jobs +40s out of a 145s build). I'd rather not keep that flag in the dev profile if it's not necessary.

tlspuffin/Cargo.toml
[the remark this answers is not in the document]
aeyno: You're right, I tried and now it works (boringssl202311 and boringssl202403 with and without asan)

Cargo.toml
aeyno: You're right, I tried and now it works (boringssl202311 and boringssl202403 with and without asan)

michaelmera (approved):

LCBH (approved):

crates/boringssl-src/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        boring_conf
+            .cflag("-fsanitize=address")
+            .cflag("-shared-libsan")
+            .cflag("-Wno-unused-command-line-argument")
+            .cflag(format!("-Wl,-rpath={}/lib/linux/", clang))
+            .cxxflag("-fsanitize=address")
+            .cxxflag("-shared-libsan")
+            .cxxflag("-Wno-unused-command-line-argument")
+            .cxxflag(format!("-Wl,-rpath={}/lib/linux/", clang));
+    }
```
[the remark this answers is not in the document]
maxammann: https://github.com/tlspuffin/tlspuffin/pull/302

crates/openssl-src-111/src/deterministic_rand.c
```diff
@@ -2,19 +2,22 @@
 #include <openssl/rand.h>
 #include <stdlib.h>
 
-static uint64_t seed = 42;
+unsigned int tlspuffin_seed = 42;
+const unsigned int m = 0xFFFFFFFF;
+const unsigned int a = 22695477;
+const unsigned int c = 1;
```
maxammann, line 8: These are no longer used. Likely a merge fault.

crates/openssl-src-111/src/openssl.rs
```diff
@@ -134,6 +147,12 @@ impl Build {
 
         clone_repo(&inner_dir.to_str().unwrap()).unwrap();
 
+        if cfg!(feature = "openssl312") {
+            let source_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+            // Patching the repo to get deterministic timestamp
+            patch_openssl(out_dir, "det_timestamp.patch").unwrap();
+        }
```
maxammann, line 154: This seems kind of dangerous to enable by default

puffin/src/trace.rs
```diff
@@ -110,6 +112,32 @@ pub struct TraceContext<PB: ProtocolBehavior + 'static> {
...
+
+impl<PB: ProtocolBehavior + PartialEq> PartialEq for TraceContext<PB> {
+    fn eq(&self, other: &Self) -> bool {
+        self.agents == other.agents
+            && self.put_registry == other.put_registry
+            && self.deterministic_put == other.deterministic_put
+            && self.default_put_options == other.default_put_options
+            && self.non_default_put_descriptors == other.non_default_put_descriptors
+            && format!("{:?}", self.knowledge) == format!("{:?}", other.knowledge)
+            && format!("{:?}", self.claims) == format!("{:?}", other.claims)
```
maxammann, line 137: This seems a bit hacky. Can we derive PartialEq?

tlspuffin/src/boringssl/deterministic.rs
```diff
@@ -0,0 +1,42 @@
...
+        boringssl::deterministic::reset_rand,
+        put_registry::TLS_PUT_REGISTRY,
+        tls::{
+            seeds::{create_corpus, seed_client_attacker_full_boring},
+            trace_helper::TraceHelper,
+        },
+    };
+
+    // BUG: This test only works in a single threaded cargo test execution
+    #[ignore]
```
maxammann, line 29: Add a TODO comment here, as we use that one typically in the codebase to mark todos.

Maybe this is even worth a task on github.

tlspuffin/src/boringssl/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+            self.stream.get_mut(),
+            result,
+        )
+    }
+
+    fn take_message_from_outbound(
+        &mut self,
+    ) -> Result<Option<MessageResult<Message, OpaqueMessage>>, Error> {
+        let memory_stream = self.stream.get_mut();
+        //memory_stream.take_message_from_outbound()
```
maxammann, line 131:
```suggestion
```

tlspuffin/src/integration_tests/determinism.rs
```diff
@@ -0,0 +1,58 @@
...
+    tls::{seeds::seed_client_attacker_full, trace_helper::TraceHelper},
+};
+
+#[test]
+#[cfg(all(
+    feature = "deterministic",
+    feature = "boringssl-binding",
+    feature = "tls13",
+    feature = "TODO"
+))] // TODO: only passes in mono-thread!! with option `-test-threads=1`
```
maxammann, line 14: Should we ignore this test?

tlspuffin/src/integration_tests/determinism.rs
```diff
@@ -0,0 +1,58 @@
...
+    put_registry::TLS_PUT_REGISTRY,
+    tls::{seeds::seed_client_attacker_full, trace_helper::TraceHelper},
+};
+
+#[test]
+#[cfg(all(
+    feature = "deterministic",
+    feature = "boringssl-binding",
+    feature = "tls13",
+    feature = "TODO"
```
maxammann, line 13:
```suggestion
```

tlspuffin/src/openssl/deterministic.rs
```diff
@@ -1,35 +1,53 @@
...
+    fn test_openssl_no_randomness_simple() {
+        determinism_set_reseed_openssl();
+        assert_eq!(get_seed(), 42);
         let mut buf1 = [0; 2];
         rand_bytes(&mut buf1).unwrap();
-        assert_eq!(buf1, [183, 96]);
+        assert_eq!(buf1, [177, 180]);
+        assert_ne!(get_seed(), 42);
+        determinism_set_reseed_openssl();
+        assert_eq!(get_seed(), 42);
```
maxammann, line 51: Why did that change?

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -79,6 +80,115 @@ pub fn fn_decrypt_handshake(
...
+
+pub fn fn_find_server_finished(messages: &Vec<Message>) -> Result<Message, FnError> {
+    for msg in messages {
+        if let MessagePayload::Handshake(x) = &msg.payload {
+            if x.typ == HandshakeType::Finished {
+                return Ok(msg.clone());
+            }
+        }
+    }
+    Err(FnError::Unknown("no finished".to_owned()))
```
maxammann, line 189: More major discussion: Does it make sense to add one function symbol instead that extracts a message by type (e.g. `HandshakeType::Finished`) or index?

tlspuffin/src/tls/rustls/msgs/message.rs
```diff
@@ -64,6 +64,43 @@ impl MessagePayload {
...
+    /// Extract multiple messages payloads from one ApplicationData message
+    pub fn multiple_new(
+        typ: ContentType,
+        vers: ProtocolVersion,
+        payload: Payload,
+    ) -> Result<Vec<Self>, Error> {
+        let fallback_payload = payload.clone();
+        let mut r = Reader::init(&payload.0);
+        let mut parsed: Vec<Self> = vec![];
+        while r.any_left() {
```
maxammann, line 76: This is pretty crazy, did not know this. Can we maybe turn that behavior off in boringssl? If not should we maybe rework more of the algebra here to model the real TLS here that can include multiple messages per ApplicationData?

maxammann (changes requested):

tlspuffin/src/tls/rustls/msgs/message.rs
maxammann: This line here indicates that plaintext only contains a single message: 

https://datatracker.ietf.org/doc/html/rfc8446#section-5.2:~:text=content%3A%20%20The%20TLSPlaintext.fragment%20value%2C%20containing%20the%20byte%0A%20%20%20%20%20%20encoding%20of%20a%20handshake%20or%20an%20alert%20message%2C%20or%20the%20raw%20bytes%20of%0A%20%20%20%20%20%20the%20application%27s%20data%20to%20send.

maxammann: We likely can switch to NULL cipher in boringssl: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/ssl/s3_both.cc#189

null ciphers should be available with boringssl: https://github.com/google/boringssl/blob/master/FUZZING.md#fuzzer-mode

tlspuffin/src/tls/seeds.rs
```diff
@@ -1822,8 +2157,10 @@ pub fn create_corpus() -> Vec<(Trace<TlsQueryMatcher>, &'static str)> {
         seed_successful12_with_tickets: cfg(feature = "tls12-session-resumption"),
         // Client Attackers
         seed_client_attacker: cfg(feature = "tls13"),
-        seed_client_attacker_auth: cfg(all(feature = "tls13", feature = "client-authentication-transcript-extraction")),
-        seed_client_attacker_full: cfg(feature = "tls13"),
+        seed_client_attacker_full: cfg(all(feature = "tls13", not(feature = "boringssl-binding"))),
+        seed_client_attacker_full_boring: cfg(all(feature = "tls13", feature = "boringssl-binding")),
+        seed_client_attacker_auth: cfg(all(feature = "tls13", feature = "client-authentication-transcript-extraction", not(feature = "boringssl-binding"))),
+        seed_client_attacker_auth_boring: cfg(all(feature = "tls13", feature = "client-authentication-transcript-extraction", feature = "boringssl-binding")),
```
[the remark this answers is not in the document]
LCBH: A bit more context on this: BoringSSL does not fragment ApplicationData the same way OpenSSL/WolfSSL/LibreSSL does. Simply put: when OpenSSL sends a flight of 4 encrypted messages, BoringSSL sends a flight of one encrypted message containing 4 application data. Both are valid per the TLS 1.3 spec, but depending on which case it is, queries and traces won't be executable.
Tom added some more symbols and adapted some seeds to be executable by BoringSSL but with the current queries, we believe there is no way to uniformly write working seeds for all PUTs. Tom is working on quite deeplu improving things regarding extracting knowledge and queries.

In the meantime, we'd like to merge BoringSSL even though not all seeds can be executed, which will be resolved later with the new queries design.

tlspuffin/src/tls/seeds.rs
```diff
@@ -1603,6 +1754,190 @@ pub fn _seed_client_attacker_full(
     )
 }
 
+pub fn seed_client_attacker_full_boring(server: AgentName) -> Trace<TlsQueryMatcher> {
```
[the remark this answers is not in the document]
LCBH: Same as above.

tlspuffin/src/wolfssl/mod.rs
```diff
@@ -263,9 +273,9 @@ impl Put<TLSProtocolBehavior> for WolfSSL {
         unsafe { version().to_string() }
     }
 
-    fn set_deterministic(&mut self) -> Result<(), puffin::error::Error> {
```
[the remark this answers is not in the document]
michaelmera: Yes. I noticed the change and we discussed it offline. I will rebase my PR while taking this into account. Not a big deal since we don't provide an actual implementation of the C PUT, for now...

crates/boringssl-src/patches/reset_drbg.patch
```diff
@@ -0,0 +1,42 @@
...
++}
++
++void RAND_reset_for_fuzzing(void) {
++  g_num_calls = 0;
++  struct rand_thread_state *state =
++      CRYPTO_get_thread_local(OPENSSL_THREAD_LOCAL_RAND);
++  rand_thread_state_free(state);
++
++  CRYPTO_set_thread_local(OPENSSL_THREAD_LOCAL_RAND, NULL, NULL);
++}
```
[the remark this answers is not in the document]
aeyno: Yes this patch works. Currently, the determinism test for BoringSSL (`test_boringssl_no_randomness_full` in `tlspuffin/src/boringssl/deterministic.rs`) is disabled due to a problem with multithreaded tests but in single thread mode this test can pass only with this patch applied.

crates/boring/test/rsa-encrypted.pem
```diff
@@ -0,0 +1,30 @@
+-----BEGIN RSA PRIVATE KEY-----
```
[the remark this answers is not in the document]
aeyno: We are not executing those tests, I will remove the files.

crates/boringssl-src/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        .define("BUILD_SHARED_LIBS", "OFF")
+        .define("CMAKE_INSTALL_PREFIX", dest.as_ref().to_str().unwrap())
+        .define("CMAKE_C_COMPILER", "clang")
+        .define("CMAKE_CXX_COMPILER", "clang++")
+        .pic(true)
+        .cflag("-g")
+        .cxxflag("-g")
+        .define("CMAKE_BUILD_TYPE", "Release")
+        .define("OPENSSL_NO_BUF_FREELISTS", "1")
+        .define("OPENSSL_NO_ASM", "1");
```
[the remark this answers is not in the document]
aeyno: I removed the flag to allow building with ASM

crates/boringssl-src/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    if options.deterministic {
+        boring_conf.define("FUZZ", "1");
+        boring_conf.define("NO_FUZZER_MODE", "1");
```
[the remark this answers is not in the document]
aeyno: I added a comment and a link to the official doc of those flags

crates/boringssl-src/src/main.rs
```diff
@@ -0,0 +1,29 @@
+use std::{env, path::PathBuf};
+
+use boringssl_src::{build, BoringSSLOptions, GitRef};
+
+fn main() {
+    // let out_dir = PathBuf::from("out_dir");
+    let out_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap()).join("out");
+
+    env::set_var("TARGET", "aarch64-apple-darwin");
+    env::set_var("HOST", "aarch64-apple-darwin");
```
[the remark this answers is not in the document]
aeyno: Yes I don't know why those lines are here. I think that at the time I just copied what was in `crates/wolfssl-src/src/main.rs` to mimic the build process from WolfSSL

crates/boringssl-src/wrapper.h
```diff
@@ -0,0 +1,3 @@
+#include <wolfssl/options.h>
+#include <wolfssl/openssl/ssl.h>
+#include <wolfssl/ssl.h>
```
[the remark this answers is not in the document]
aeyno: Done

crates/boringssl-src/src/lib.rs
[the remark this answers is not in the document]
michaelmera: Yes, we had a discussion and concluded that it's not worth spending the time to fix now. We would need to duplicate the code I used in #302 or extract it in a common crate. This will be comparatively easy to fix when we make the move to building vendor libraries outside of cargo. So I support the idea that this can remain broken on macos until then.

crates/boringssl-sys/build.rs
```diff
@@ -0,0 +1,124 @@
...
+    })
+    .unwrap();
+
+    // Linking Time!
+    // The parameters to link BoringSSL are inspired from
+    // https://github.com/cloudflare/boring/blob/master/boring-sys/build/main.rs
+    println!("cargo:rustc-link-search={}/lib", out_dir.display());
+    println!("cargo:rustc-link-lib=static=crypto");
+    println!("cargo:rustc-link-lib=static=ssl");
+    println!("cargo:rustc-link-lib=stdc++");
```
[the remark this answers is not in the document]
maxammann: This needs to be addressed as discussed in Mattermost. This is causing issues with Rust linking.

tlspuffin/src/tls/rustls/msgs/message.rs
aeyno: > This line here indicates that plaintext only contains a single message:
> 
> https://datatracker.ietf.org/doc/html/rfc8446#section-5.2:~:text=content%3A%20%20The%20TLSPlaintext.fragment%20value%2C%20containing%20the%20byte%0A%20%20%20%20%20%20encoding%20of%20a%20handshake%20or%20an%20alert%20message%2C%20or%20the%20raw%20bytes%20of%0A%20%20%20%20%20%20the%20application%27s%20data%20to%20send.

From section 5.1 : 
```
 Handshake messages MAY be coalesced into a single TLSPlaintext record
   or fragmented across several records, provided that:

   -  Handshake messages MUST NOT be interleaved with other record
      types.  That is, if a handshake message is split over two or more
      records, there MUST NOT be any other records between them.

   -  Handshake messages MUST NOT span key changes.  Implementations
      MUST verify that all messages immediately preceding a key change
      align with a record boundary; if not, then they MUST terminate the
      connection with an "unexpected_message" alert.  Because the
      ClientHello, EndOfEarlyData, ServerHello, Finished, and KeyUpdate
      messages can immediately precede a key change, implementations
      MUST send these messages in alignment with a record boundary.
```

crates/boringssl-sys/build.rs
```diff
@@ -0,0 +1,124 @@
+use std::{env, path::PathBuf};
+
+use boringssl_src::{build, BoringSSLOptions, GitRef};
+
+fn main() {
+    let git_ref: GitRef = if cfg!(feature = "boring-2023-11-18") {
+        GitRef::Commit(String::from("698aa894c96412d4df20e2bb031d9eb9c9d5919a"))
+    } else if cfg!(feature = "boring-2024-03-22") {
+        GitRef::Commit(String::from("368d0d87d0bd00f8227f74ce18e8e4384eaf6afa"))
+    } else if cfg!(feature = "vendored-master") {
```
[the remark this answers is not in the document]
aeyno: BoringSSL doesn't have a release mechanism, every Google project that uses BoringSSL just build on main, so I've selected two commits where we know that the PUT build and that we can use for testing. For the moment it should be able to build on main, until one commit changes the API or the build process...

tlspuffin/src/tls/fn_utils.rs
aeyno: We are thinking about this for a next PR (probably along the rework of the queries). For now, we are going to keep things like this for this PR

tlspuffin/src/boringssl/deterministic.rs
aeyno: Done

crates/openssl-src-111/src/deterministic_rand.c
aeyno: Yes, I removed those lines

crates/openssl-src-111/src/openssl.rs
LCBH: Will be removed. That was an attempt at making OpenSSL fully deterministic, even the timestamps. DId not work in the end.

puffin/src/trace.rs
LCBH: We do not impose equality of all fields. We could have used the [derivative](https://crates.io/crates/derivative) to exclude those but I didn't know this crate at the time. Plus, this is solely used for testing the determinism.

tlspuffin/src/boringssl/mod.rs
LCBH: Will be pushed.

tlspuffin/src/integration_tests/determinism.rs
LCBH: We'd like to keep this test to be run manually in mono-thread mode. We cannot include it non-conditionally because the CI will fail otherwise as it is not in mono thread. Adding a mono-thread feature does not seem ideal either.

tlspuffin/src/integration_tests/determinism.rs
LCBH: This will make the CI fail.

tlspuffin/src/openssl/deterministic.rs
LCBH: Had to do this accoridnly to the the changes made in [crates/openssl-src-111/src/deterministic_rand.c](https://github.com/tlspuffin/tlspuffin/pull/290/files/8865f8b83db41dc5cf9ac7c220f2711d0b01e8ef#diff-ff8a9ea809dc3abf285d6dfdeaf19161e1847aaf1c6102ac2c345639ba639c8f). As we revert the later, we will revert this as well.

crates/boringssl-sys/build.rs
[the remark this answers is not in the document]
LCBH: Will be documented in the README as such.

crates/openssl-src-111/patches/det_timestamp.patch
```diff
@@ -0,0 +1,54 @@
...
++//    else
++//        r.t = ((uint64_t)t.tv_sec * 1000000 + t.tv_usec) * OSSL_TIME_US;
+ #endif  /* defined(_WIN32) */
+-    return r;
++    r = ossl_time_zero();
++    r.t = r.t + tlspuffin_time * (1000 * OSSL_TIME_US); // 1 seconds elapsed between every call
++    tlspuffin_time++;
++    return ossl_time_zero();
++//    return r;
+ }
```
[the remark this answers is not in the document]
LCBH: Removed.

crates/openssl-src-111/src/deterministic_rand.c
```diff
@@ -2,19 +2,22 @@
 #include <openssl/rand.h>
 #include <stdlib.h>
 
-static uint64_t seed = 42;
+unsigned int tlspuffin_seed = 42;
```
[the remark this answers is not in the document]
LCBH: Reverted.

LCBH:
Good to merge :+1:

tlspuffin/src/tls/rustls/msgs/message.rs
maxammann: Sorry, I was not very clear. The code here suggests that using the NULL cipher avoid putting messages into a single record: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/ssl/s3_both.cc#189

(a small patch would likely also make it work without null ciphers)

PR #291, opened 2023-12-13
Fixing implicit-function-declaration errors during OpenSSL compilation

aeyno:
Fixing implicit-function-declaration errors during OpenSSL111 build which occurs with Clang 15+ (https://github.com/llvm/llvm-project/commit/7d644e1215b376ec5e915df9ea2eeb56e2d94626)

### DIFF
--- a/crates/openssl-src-111/src/lib.rs
+++ b/crates/openssl-src-111/src/lib.rs
@@ -188,6 +188,15 @@ impl Build {
 
         configure.arg("-fPIE"); // -fPIC was previously added through Cargo flags
 
+        // Recent versions of Clang returns implicit-function-declaration
+        // errors when building OpenSSL111
+        #[cfg(any(
+            feature = "openssl111k",
+            feature = "openssl111j",
+            feature = "openssl111u"
+        ))]
+        cflags.push_str(" -Wno-implicit-function-declaration ");
+
         if cfg!(feature = "sancov") {
             cflags.push_str(" -fsanitize-coverage=trace-pc-guard ");
         }


maxammann:
How would you feel about making Clang a proper dependency by adapting Nix?
Over the years this will become a catchup game and already has been.

LCBH:
> How would you feel about making Clang a proper dependency by adapting Nix? Over the years this will become a catchup game and already has been.

That would be useful indeed. Same applies for `llvm`, not shipped with `libFuzzer` on Apple's XCode for example.

In the meantime, I will merge this.

michaelmera:
My 2 cents before this get merged.

From a quick test (clang 17.0.6 on branch fuzz-OpenSSL_1_1_1j), the code preventing the build is entirely our own (`fill_claim` in `openssl/ssl/statem/statem.c`).

There are adverse effects to an implicit function declaration, namely the compiler generates its own signature (`int fill_claim();`) and calling a function with a non-matching signature [is undefined behavior](https://wiki.sei.cmu.edu/confluence/display/c/EXP37-C.+Call+functions+with+the+correct+number+and+type+of+arguments) if we provide wrong arguments. I'd rather fix the root cause of the warning by declaring the function in that file and not ignore all those errors in the future, risking strange crashes on more exotic compilers and architectures.

Additionally, the build system will be re-written for these external PUT dependencies, so I'll have the same problem anyway in a few weeks if this is not fixed on every branch where we added claims.

LCBH:
Totally 

> My 2 cents before this get merged.
> 
> From a quick test (clang 17.0.6 on branch fuzz-OpenSSL_1_1_1j), the code preventing the build is entirely our own (`fill_claim` in `openssl/ssl/statem/statem.c`).
> 
> There are adverse effects to an implicit function declaration, namely the compiler generates its own signature (`int fill_claim();`) and calling a function with a non-matching signature [is undefined behavior](https://wiki.sei.cmu.edu/confluence/display/c/EXP37-C.+Call+functions+with+the+correct+number+and+type+of+arguments) if we provide wrong arguments. I'd rather fix the root cause of the warning by declaring the function in that file and not ignore all those errors in the future, risking strange crashes on more exotic compilers and architectures.
> 
> Additionally, the build system will be re-written for these external PUT dependencies, so I'll have the same problem anyway in a few weeks if this is not fixed on every branch where we added claims.

Totally agree, we should fix `openssl/ssl/statem/statem.c` in our openssl clone. We'll open a PR for that.

michaelmera:
> Totally agree, we should fix `openssl/ssl/statem/statem.c` in our openssl clone. We'll open a PR for that.

I went through the different branches in our tlspuffin/openssl fork. Here is the corresponding status for this particular issue:

- [ ] branch fuzz-OpenSSL_1_1_1k: broken, fixed by tlspuffin/openssl#1
- [ ] branch fuzz-OpenSSL_1_1_1j: broken, fixed by tlspuffin/openssl#2
- [ ] branch fuzz-OpenSSL_1_1_1u: broken, fixed by tlspuffin/openssl#3
- [X] branch fuzz-OpenSSL_1_0_1j not implementing claims
- [X] branch fuzz-OpenSSL_3_1_2 was already fixed
- [X] branch fuzz-OpenSSL_3_1_2-time was already fixed
- [X] branch fuzz-OpenSSL_3_2 was already fixed
- [ ] branch skip-transcript: broken, fixed by tlspuffin/openssl#5
- [X] branch master not implementing claims

If no one has any objection, I will merge the pull requests tomorrow, as I need this for rebasing #297 properly. I will close this one once this is done.

**Note** There are other build issues on some of these branches but the PRs linked here only address the implicit function declarations. I was able to build and run all the branches but if you also want to do so, you might need to address these issues yourself. More PRs for those are in the making.

PR #292, opened 2023-12-13
Adding client-attacker-full trace to the corpus of traces for tls13

aeyno:
Adding client-attacker-full trace to the corpus of traces for tls13


LCBH:
Will merge, adding this trace could lead to finding new interesting test cases

PR #293, opened 2023-12-14
openssl deterministic test fails

LCBH:
- Implementing `PartialEq` and `Debug` for `TraceContext` so that we can compare the two `TraceContext` obtained at different times from executing the same trace.
- Using that to test PUT's determinism after executing `client_attacker_full`.
- Currently fails for OpenSSL 111j. See the Issue #294 for this new failing uni test `openssl::deterministic::tests::test_openssl_no_randomness_full`.

### DIFF
--- a/crates/openssl-src-111/patches/det_timestamp.patch
+++ b/crates/openssl-src-111/patches/det_timestamp.patch
@@ -1,0 +1,54 @@
+Subject: [PATCH] deterministic timestamp
+---
+Index: crypto/time.c
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/crypto/time.c b/crypto/time.c
+--- a/crypto/time.c	(revision f925bfebbb287321133b9251e72bee869a0f58b4)
++++ b/crypto/time.c	(date 1703166565605)
+@@ -11,6 +11,9 @@
+ #include <openssl/err.h>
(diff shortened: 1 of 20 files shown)


LCBH:
Commit [#aa340](https://github.com/tlspuffin/tlspuffin/pull/293/commits/aa340e58029748293e4d9c92a357460d2ab97d04) fixed the issue:
> New determinism architecture:
>  - distinguish re-set RAND from reseed
>  - reset RAND once for all for all factories with `determinism_set_reseed_all_factories`
>  - reseed all factories before executing a trace with `determinism_reseed_all_factories`
>  --> New integration test for determinism. Passes for OpenSSL!

The test `test_attacker_full_det_recreate` now passes for `openssl111j`.

LCBH:
Conclusion: this test fails because of session tickets sent after the final Server FInishes. Those contain time-related information, which thus differ across executions.

In BoringSSL, the flag `BORINGSSL_UNSAFE_DETERMINISTIC_MODE` reseed the PRNG but also make BoringSSL use a hard-coded time instead of the actual time. We have not found the equivalent for OpenSSL.
A solution would be to use [`libfaketime`](https://github.com/wolfcw/libfaketime/blob/master/README#L414) but this will also impact tlspuffin's timestamps.
One could also hijack`ossl_time_now` but that's not really robust in the long term.

LCBH:
Will close without merging, the useful more complete deterministic uni test is also in #290 that will me merged.

PR #300, opened 2024-02-27
 PUT rearchitecture: factor out OpenSSL vendored build

michaelmera:
**NOTE:** This PR is an attempt to restrict the scope of PR #282 by extracting the commits laying the groundwork for the architectural changes.

This PR introduces a new tool `mk_vendor` to build vendor libraries (OpenSSL, WolfSSL, ...) independently from Rust and cargo. It also replaces the internals of the `openssl-src` crate with calls to this tool.

In a next PR, we will introduce the API for PUT interfaces written in C. Having first this independent build system for the vendor libraries, the Rust and C PUT implementations will be able to coexist and use the same underlying vendor library. This gives us a common ground to compare and validate the new C implementation.

From a longer-term perspective, `mk_vendor` provides more flexibility and a finer control over the OpenSSL build compared to the preset cargo features. This should make it easier in the future to configure, fuzz and compare a wider range of vendor versions.

### DIFF
--- a/puts/vendor/configs/openssl.toml
+++ b/puts/vendor/configs/openssl.toml
@@ -2,32 +2,32 @@
 default = "openssl312"
 
 [preset.openssl312]
-name = "openssl-3.1.2"
+name = "openssl312"
 fetch = "openssl312"
 build = ["base", "parallel"]
 
 [preset.openssl111k]
-name = "openssl-1.1.1k"
+name = "openssl111k"
 fetch = "openssl111k"
 build = ["base", "parallel"]
 
 [preset.openssl111j]
-name = "openssl-1.1.1j"
+name = "openssl111j"
 fetch = "openssl111j"
 build = ["base", "parallel"]
 
 [preset.openssl111u]
-name = "openssl-1.1.1u"
+name = "openssl111u"
 fetch = "openssl111u"
 build = ["base", "parallel"]
 
 [preset.openssl101f]
-name = "openssl-1.0.1f"
+name = "openssl101f"
 fetch = "openssl101f"
 build = ["base"]
 
 [preset.openssl102u]
-name = "openssl-1.0.2u"
+name = "openssl102u"
 fetch = "openssl102u"
 build = ["base"]
 
--- a/puts/vendor/configs/openssl.toml
+++ b/puts/vendor/configs/openssl.toml
@@ -1,0 +1,52 @@
+[preset]
+default = "openssl312"
+
+[preset.openssl312]
+name = "openssl-3.1.2"
+fetch = "openssl312"
+build = ["base", "parallel"]
+
+[preset.openssl111k]
+name = "openssl-1.1.1k"
+fetch = "openssl111k"
+build = ["base", "parallel"]
+
+[preset.openssl111j]
+name = "openssl-1.1.1j"
+fetch = "openssl111j"
+build = ["base", "parallel"]
+
+[preset.openssl111u]
+name = "openssl-1.1.1u"
+fetch = "openssl111u"
+build = ["base", "parallel"]
+
+[preset.openssl101f]
+name = "openssl-1.0.1f"
+fetch = "openssl101f"
+build = ["base"]
+
+[preset.openssl102u]
+name = "openssl-1.0.2u"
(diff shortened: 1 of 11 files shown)


maxammann:
Note that in PRs we currently don't built all fuzzer fuzzers. For major changes it might be good to enable the ci jobs for all.

michaelmera:
> Note that in PRs we currently don't built all fuzzer fuzzers. For major changes it might be good to enable the ci jobs for all.

Good point! I will do that.

This is still work in progress but in the meantime I've been playing around with different configurations and environments. I would also like to land or drop #234 before merging this.

puts/vendor/configs/openssl.toml
```diff
@@ -0,0 +1,52 @@
...
+openssl111j = { git = "https://github.com/tlspuffin/openssl", ref = "fuzz-OpenSSL_1_1_1j" }
+openssl111u = { git = "https://github.com/tlspuffin/openssl", ref = "fuzz-OpenSSL_1_1_1u" }
+openssl101f = { git = "https://github.com/tlspuffin/openssl", ref = "OpenSSL_1_0_1f" }
+openssl102u = { git = "https://github.com/tlspuffin/openssl", ref = "OpenSSL_1_0_2u" }
+
+[patch]
+
+[build.base]
+cmake = "openssl"
+options = ["asan", "sancov", "gcov", "llvm_cov"]
```
LCBH, line 46: Don't we want to have asan as opt-in? With something like `[build.asan]` and then we can pick up a preset and add `asan` when we want to compile this preset but with Asan enabled.

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+NAME="mk_vendor"
+VERSION="0.1"
+
+USAGE="Usage:
```
LCBH, line 8: Would that make sense to extract this to a `tools/README` and link that from the main `README` ?

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
...
+
+Common Options:
+    -h, --help  display help and exit
+    --version   display version and exit
+
+Examples:
+    build the default version of OpenSSL:
+      \$ ${NAME} make openssl
+
+    build OpenSSL 1.1.1u:
```
LCBH, line 30: specify this is a preset? E.g.,:
> build OpenSSL 1.1.1u (as defined by the `preset` openssl111u):

LCBH (changes requested):

puts/vendor/configs/openssl.toml
michaelmera: The `options` in the `.toml` declare the options but they are not set by default. You add `--options=asan` to build a given preset with asan, e.g.:
```sh
$ mk_vendor openssl:openssl111k --options=sancov,asan
```

So I think this is working as you expect. Or am I not understanding what you are asking for?

tools/mk_vendor
michaelmera: I would rather create a specific documentation in the tlspuffin book.

At some point, `mk_vendor` should be rewritten in rust/python so that we don't have to maintain a big chunk of bash script. Then it will be easier to do link documentations together...

tools/mk_vendor
michaelmera: I am a bit unsure about this.

In my opinion, the help text is not there to document the tool, but rather provide a quick reminder of the commands and options. I feel like if we start talking about presets here, we should explain what it is in more details, and that the `openssl` part matches the file `puts/vendor/configs/openssl.toml`, and that ...

I think I would rather develop a more complete documentation separated from the help text.

LCBH: OK if this is documented elsewhere.

tools/mk_vendor
LCBH: Yes, this will obviously be resolved with the future documentation.

puts/vendor/configs/openssl.toml
LCBH: RIght! I got confused ;)

LCBH (approved):

maxammann (approved):

PR #301, opened 2024-03-04
improve CI actions

michaelmera:
**WARNING**: The default github actions permissions for the entire tlspuffin organization have been set to "restricted" mode. If you came here because a previously working workflow now fails because of insufficient permissions, set the required permissions explicitly in this particular workflow. See the [github documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication?ref=blog.gitguardian.com#permissions-for-the-github_token) for the list of default permissions granted in "restricted" mode.

This PR aims to provide several improvements over the current github actions:
- add justfile recipe to run github actions locally
- fix calls to [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache)
- update external actions relying on deprecated Node 16
- add ability to control the CI workflow based on PR labels
- add smoke test before running the complete build matrix
- fix CI/CD permissions

See also:
- Github's [notice of deprecation for Node 16](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)
- Github's [introduction to workflow permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)

**PENDING** 
- [x] Waiting for #304 to land because the new CI relies on some changes/fixes included in that PR.
- [ ] The following external action still need to update to Node 20: peaceiris/actions-mdbook#500


.github/workflows/ci.yml
```diff
@@ -0,0 +1,176 @@
...
+
+  test-tlspuffin:
+    needs: [configure, smoke-test]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: openssl111
+            features: openssl111
+            save-cache: "true" # We only save the cache once, else we get too mache cache entries
```
LCBH, line 101: typo

LCBH (changes requested):
Seems like CI is now disabled on PRs forked from a branch different from `main`, despite using the `ci:full` flag. Would be nice to keep the control on when CI is enabled with the flags.

puffin/src/cli.rs
```diff
@@ -207,6 +224,42 @@ pub fn main<PB: ProtocolBehavior + Clone + 'static>(
...
+                }
+            })
+            .collect::<Vec<_>>();
+
+        paths.sort_by_key(|path| {
+            fs::metadata(path)
+                .unwrap_or_else(|_| panic!("missing trace file {}", path.display()))
+                .modified()
+                .unwrap()
+        });
```
maxammann, line 253: Could be de-duplicated, left that comment in a different PR already.

tlspuffin/src/openssl/mod.rs
```diff
@@ -135,7 +130,23 @@ impl Put<TLSProtocolBehavior> for OpenSSL {
...
+            // FIXME don't force-register a new claimer on reset
+            //
+            //    Because OpenSSL vendor libraries crash when no claimer is
+            //    registered (#253), we are forced to register a new one on
+            //    reset. This should be removed once this bug is fixed.
+            //
+            //    See the WolfSSL module for comparison.
+            #[cfg(feature = "claims")]
+            self.register_claimer(self.config.descriptor.name);
+        }
```
maxammann, line 149: Feels like some changes leaked into this PR.

test_tlspuffin.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
maxammann, line 1: Is this file used anywhere?

justfile
```diff
@@ -52,3 +110,50 @@ clear-gh-caches:
...
+#   $ just act --list
+#
+# - do a dry-run:
+#   $ just act -n
+#
+# - run only the configure job:
+#   $ just act -j configure
+#
+# - emulate a pull_request against the main branch:
+#   $ just ACT_EVENT_NAME=pull_request ACT_EVENT_FILE=${PWD}/.github/act/events/pr_to_main.json act
```
maxammann, line 130: I think it would be great to have a reference tot his in the README

justfile
```diff
@@ -3,18 +3,37 @@
...
+check-workspace: install-clippy
+  cargo clippy
+
+# run clippy on a vendored crate (e.g. libressl-src)
+check-crate NAME FEATURES: install-clippy
+  #!/usr/bin/env bash
+  cleanup() {
+    find {{ justfile_directory() / "crates" }} -name Cargo.lock -exec rm -f '{}' +
+  }
+  trap cleanup EXIT
```
maxammann, line 32: I think this is a little dangerous to have in a file that can be executed by users. I maybe don't want to remove my lock file when running a check locally.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    -   id: check-vcs-permalinks
+    -   id: destroyed-symlinks
+    -   id: detect-private-key
+    -   id: fix-byte-order-marker
+    -   id: mixed-line-ending
+
+-   repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+    rev: v2.12.0
+    hooks:
+    -   id: pretty-format-rust
```
maxammann, line 21: Not sure how this works so it would be nice to mention this in the README

.github/workflows/run-checks.yml
```diff
@@ -1,16 +1,58 @@
...
+      - name: check `libressl-src`
+        run: just check-crate libressl-src libressl333
+      - name: check `libressl-src`
+        run: just check-crate libressl-src libressl333
+      - name: check `openssl-src`
+        run: just check-crate openssl-src-111 openssl101f
+      - name: check `wolfssl`
+        run: just check-crate wolfssl wolfssl430,wolfssl-sys/wolfssl430
+      - name: check `wolfssl-sys`
+        run: just check-crate wolfssl-sys wolfssl540,fix-CVE-2022-39173,fix-CVE-2022-42905
```
maxammann, line 58: Doesn't this take a long time to execute? I typically try to run cargo checks directly after build as then it might reuse build artifacts from the build step.

.github/act/events/pr_to_main.json
```diff
@@ -0,0 +1,10 @@
+{
+    "pull_request": {
+        "head": {
+            "ref": "build-vendor"
+        },
+        "base": {
+            "ref": "main"
+        }
+    }
+}
```
maxammann, line 10: What are these files in act/events used for?

.github/workflows/on_main_push.yml
```diff
@@ -1,137 +0,0 @@
...
-          - wolfssl540
-          - wolfssl540-sdos2
-          - wolfssl530-cdos
-          - wolfssl540_asan
-          - wolfssl540-buf
-          - wolfssl540-heap
-          - wolfssl540_asan-perf
-          - wolfssl540-perf
-          - libssh
-          - tlspuffin_no_default_features
```
maxammann, line 46: It seems like we are no longer building the tlspuffin binaries in CI or did I miss a file?

Building binaries is important additionally to tests to catch issues with that.

maxammann (commented):
Feels like some changes from other PRs leaked into this PR.

.github/workflows/ci.yml
michaelmera: Not mine though 😆

@maxammann I'm not sure if it was meant to say that we get _too many_ incompatible cache entries OR that incompatible cache entries were _mashed_ together, resulting in rebuild. Anyway, the caching strategy will change once we can do vendored library builds outside of cargo, through `mk_vendor`. So I guess it will be the right time to remove this comment.

puffin/src/cli.rs
michaelmera: Yes, I forgot to mention that this PR is stacked on #304. Sorry... 😢

justfile
michaelmera: Yes, good point. I have taken note to work on this when I do the PR related to tlspuffin's documentation.

.pre-commit-config.yaml
michaelmera: Oops. This should be here. Those are my local pre-commit checks. I don't think we plan to use `pre-commit` for this project, so I will remove this file 😅

test_tlspuffin.sh
michaelmera: Nope. I was apparently too eager to add files in my commits. This and `test_mk_vendor.sh` are local scripts of mine that should be removed. I'll revert that.

.github/workflows/run-checks.yml
michaelmera: Not really. Because these are built in the same job and most dependencies are shared, it doesn't take that much time. Also, now that we have this tested independently I removed the (mostly duplicated) checks step in the extended build matrix, so we gain a bit of time there.

.github/workflows/on_main_push.yml
michaelmera: No, we still build them, but the entry point is now in the `ci.yml` workflow. The test matrix has not changed in this PR.

.github/act/events/pr_to_main.json
michaelmera: These files are placeholder events to simulate the events received by the github CI, so that we can run the CI workflows locally with [act](https://github.com/nektos/act). It's a project to run github actions locally through docker. It has many flaws in my opinion but that's the best local alternative I could find for quickly testing my changes while developing this PR.

michaelmera:
> Seems like CI is now disabled on PRs forked from a branch different from `main`, despite using the `ci:full` flag. Would be nice to keep the control on when CI is enabled with the flags.

@LCBH I'm not sure what you mean. This PR makes CI run a reduced set of checks on every PR, regardless of the source branch and source repository. The label `ci:full` activates the full set of CI checks on the annotated PR. The full set of checks always run when pushing to the branches `main` and `trailofbits`. I didn't change the behavior in that respect. Did I miss something?

LCBH (approved):

.github/workflows/ci.yml
maxammann: Not sure anymore sorry

PR #302, opened 2024-03-06
fix macos builds with asan

michaelmera:
This PR fixes the rpath when building tlspuffin with address sanitizer on macos.

We should now be able to:
- run asan-based tests on local macos machines
- unify the CI process and include macos runners in github actions (see also #301)

Note that for older compilers that don't support `--print-runtime-dir` we still rely on constructing the compiler-rt runtime path manually (as it was the case on Linux before this fix).


maxammann (commented):
Looks all good here!

PR #303, opened 2024-03-08
fix build when incompatible features are requested

michaelmera:
Right now, most vendored crates are part of the workspace and are therefore built by default when no specific packages is specified on the `cargo` command line. Not only does that lead to additional build time, but the default features of these crates don't fit what is requested when they are pulled in by `tlspuffin` and `sshpuffin`, so they will need to be recompiled again when used.

Additionally, incompatible vendor features defined in the `tlspuffin` crate are silently resolved and can result in one arbitrary vendor library being included in the final build, while all the capability (resumption, tls3, ...) feature flags for all the selected features are active. In the best case the build fails with error messages not pointing to the exact problem. In the worst case the build finishes but strange behavior can happen at runtime.

Example of problematic builds:
```sh
$ cargo test -p tlspuffin --features=openssl111,openssl111u
$ cargo test -p tlspuffin --features=openssl111,openssl312
```

This PR ensures that:

- we only build openssl/wolfssl/libressl/libssh vendored crates when they are explicitly requested.
- we fail the build with an appropriate compile-time error message when incompatible features are requested.

**Note** This PR is a best-effort attempt to handle mutually exclusive features through the use of conditional compilation macros and attributes. It is not ideal, as vendored dependencies (correctly configured now!) will still be built before the compilation actually stops. As long as this is not supported directly in cargo, this is probably the best we can do while preserving these features. An alternative would be to use environment variables to select the vendored versions but is more effort than I'm willing to make given that these features will disappear later anyway (see next point).

**Note** This problem will be completely solved once we build the PUT vendor libraries independently of `cargo` (#300) and we use PUT written in C as harness (#282) to pull them into `tlspuffin`.

References:
- see rust-lang/cargo#2980 and [this Pre-RFC](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618) for upstream discussion on mutually exclusive features in `cargo`.
- see [sqlx](https://github.com/launchbadge/sqlx/blob/135d16a34f2a05b075a7939d129d9076f4356c06/sqlx-rt/src/lib.rs) (similar to this PR) and [inkwell](https://github.com/TheDan64/inkwell/blob/e0aa2e51a6cb501e4e2a889bbea12a1efab9c4ff/internal_macros/src/lib.rs) (use procedural macros) for two different implementations trying to solve the same problem.


maxammann (commented):
Looks good! Yeah this will be resolved in the future!

PR #304, opened 2024-03-11
QoL changes for better CI integration

### LINKED ISSUE #154: Avoid using SSL_clear
Its non-default to use SSL_clear in order to reset a session. Allocate a new SSL session instead.

michaelmera:
This PR includes several quality of life changes for problems encountered during the CI actions update.

Mostly minor changes, but two things to take note of:
- it fixes #154  by implementing non-clear mode in the OpenSSL PUT
- ~~it changes the exit code of the `execute` subcommand~~ (see revert below)

~~I adapted the script `evaluation-paper-2023/coverage.sh` to the new `execute` exit code, so that we can rerun it with this new version of tlspuffin. @maxammann  @LCBH @aeyno Tell me if you expect this change to break any of your own scripts.~~

### DIFF
--- a/puffin/src/cli.rs
+++ b/puffin/src/cli.rs
@@ -61,7 +61,7 @@ fn create_app() -> Command {
                 .arg(arg!(--multiple "Whether we want to output multiple views, additionally to the combined view"))
                 .arg(arg!(--tree "Whether want to use tree mode in the combined view")),
             Command::new("execute")
-                .about("Executes a trace stored in a file. The exit code describes if more files are available for execution.")
+                .about("Executes a trace stored in a file.")
                 .arg(arg!(<inputs> "The file which stores a trace").num_args(1..))
                 .arg(arg!(-n --number <n> "Amount of files to execute starting at index.").value_parser(value_parser!(usize)))
                 .arg(arg!(-i --index <i> "Index of file to execute.").value_parser(value_parser!(usize)))
@@ -170,24 +170,28 @@ pub fn main<PB: ProtocolBehavior + Clone + 'static>(
 
         paths.sort_by_key(|path| fs::metadata(path).unwrap().modified().unwrap());
 
-        let mut end_reached = false;
-
         let lookup_paths = if index < paths.len() {
             if index + n < paths.len() {
                 &paths[index..index + n]
             } else {
-                end_reached = true;
                 &paths[index..]
             }
         } else {
-            end_reached = true;
             // empty
             &paths[0..0]
         };
 
+        info!("execute: found {} inputs", paths.len());
+        info!(
+            "execute: running on subset [{}..{}] ({} inputs)",
+            index,
+            index + n,
+            lookup_paths.len()
+        );
+
         for path in lookup_paths {
             info!("Executing: {}", path.display());
-            execute(&path, put_registry);
+            execute(path, put_registry);
         }
 
         if !lookup_paths.is_empty() {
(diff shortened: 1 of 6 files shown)


michaelmera:
Due to popular demand, I'll reverse the changes on the `execute` command and add a new command instead, so that we keep full compatibility with our current coverage scripts.

tlspuffin/src/openssl/mod.rs
```diff
@@ -135,7 +130,23 @@ impl Put<TLSProtocolBehavior> for OpenSSL {
...
     fn reset(&mut self, _agent_name: AgentName) -> Result<(), Error> {
-        bindings::clear(self.stream.ssl()); // FIXME: Add non-clear method like in wolfssl
+        if self.config.use_clear {
+            bindings::clear(self.stream.ssl());
+        } else {
+            self.stream = Self::new_stream(&self.ctx, &self.config).map_err(|err| {
+                Error::Put(format!("OpenSSL error during stream creation: {}", err))
+            })?;
+
+            // FIXME don't force-register a new claimer on reset
```
LCBH, line 140: It seems this modifies the behavior when using the OpenSSL vendor and when the `--put-use-clear` is not used. I think we should make `put-use-clear` the default behavior and makes it opt-out.

LCBH (dismissed):
Let's discuss the `put-use-clear` flag.

tlspuffin/src/openssl/mod.rs
maxammann: The put-use-clear was required to reproduce a specific vulnerability which only works when using clear() instead of creating a new SSL context. 

We could discuss whether we want to remove this feature and bring that back eventually. There are likely a lot of similar options we want to have at some point that affect intitialization.

puffin/src/cli.rs
```diff
@@ -208,6 +225,42 @@ pub fn main<PB: ProtocolBehavior + Clone + 'static>(
...
+                }
+            })
+            .collect::<Vec<_>>();
+
+        paths.sort_by_key(|path| {
+            fs::metadata(path)
+                .unwrap_or_else(|_| panic!("missing trace file {}", path.display()))
+                .modified()
+                .unwrap()
+        });
```
maxammann, line 254: This looks a lot like the code above. Can we deduplicate this?

tlspuffin/src/openssl/mod.rs
maxammann: Yeah this is kind of changing things, but in general, creating a new SSL struct is the right thing to do to keep things more isolated. So I would not be concerned here with changing the default.

This was more like a historic thing and changing this should not cause any problems.

maxammann (approved):

puffin/src/cli.rs
michaelmera: I added this command because I wanted something in CI to run traces and reliably return non-zero exit code when something wrong happened. This code is duplicated on purpose so that in the future we have an incentive to merge these two commands back together.

The full story if you are interested: we had a discussion offline about the current state of the `execute` command. I don't agree with the current use of batching files inside of tlspuffin and I also don't agree with ordering based on files metadata which are unreliable. We agreed to the following:
- we don't want to break the current coverage scripts;
- we need a reliable command exit code for the CI
- we don't want to spend time now to write new coverage scripts and make sure the result is consistent with previous usage.
 
So this is the compromise that makes everyone unhappy: adding a new command that is mostly similar to `execute`. At some point, we will review the way we extract coverage and we can get rid of this duplication.

tlspuffin/src/openssl/mod.rs
LCBH: There are two dimensions here:
 1. efficiency: it is expected to be faster to clear versus to recreate SSL struct. I don't have figures measuring this though. Anyone has?
 2.  fuzzing capabilities: 
    - using CLEAR allows to reach bugs e.g., based on side-conditions from previous executions impacting current executions
    - not using CEAR could potentially allow to execute more traces, some of them triggering bugs, that would otherwise be rejected when using CLEAR.
    
I don't know how to balance this trade-off now but we don't have to before merging this PR I think.

PR #305, opened 2024-03-19
PUT rearchitecture: introduce C PUT harness

michaelmera:
**NOTE:** This PR is an attempt to restrict the scope of PR #282 by extracting the commits laying the groundwork for the architectural changes.

This PR introduces the harness for PUTs written in C. This includes:
- a C harness interface mostly equivalent to the current C `puffin::Put` trait.
- a cmake script to build dummy C PUTs, linked statically to an arbitrary version of OpenSSL. 
- a rust crate `tlspuffin-harness-sys` defining the ffi calls from tlspuffin into the PUT harness written in C.
- a reviewed process for dynamically registering PUTs to a `PutRegistry` instead of the static registration we had until now.


puffin/src/trace.rs
```diff
@@ -96,13 +96,13 @@ impl<M: Matcher> fmt::Display for Knowledge<M> {
...
 /// which have need exchanged and are not yet processed by an output step.
-pub struct TraceContext<PB: ProtocolBehavior + 'static> {
+pub struct TraceContext<PB: ProtocolBehavior> {
     /// The knowledge of the attacker
     knowledge: Vec<Knowledge<PB::Matcher>>,
     agents: Vec<Agent<PB>>,
     claims: GlobalClaimList<PB::Claim>,
 
-    put_registry: &'static PutRegistry<PB>,
+    put_registry: PutRegistry<PB>,
```
maxammann, line 105: Note to myself: Why is put registry no longer required to be static?

puffin/src/protocol.rs
```diff
@@ -58,11 +57,6 @@ pub trait ProtocolBehavior: 'static {
     /// Get the signature which is used in the protocol
     fn signature() -> &'static Signature;
 
-    /// Gets the registry for concrete programs-under-test.
```
maxammann, line 61: Was this function unused?

puts/cmake/CheckCCompilerCanPrintDeps.cmake
```diff
@@ -0,0 +1,21 @@
+include_guard(GLOBAL)
```
maxammann, line 1: This file needs some comment what it is and how to use it. Not a CMake pro myself.

puts/harness/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+enable_testing()
+
+set(PUTS "")
+set(PUT_LIBS "")
+
+string(REPLACE "," ";" VENDORS "${VENDORS}")
+foreach(vendor_path IN LISTS VENDORS)
+    set(PROTOCOL "tls")
+    set(HARNESS "openssl")
```
maxammann, line 13: We need to document these parameters somewhere. Not sure what a good way would be

puts/harness/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+cmake_minimum_required(VERSION 3.15)
+
+project(puts VERSION 0.1 LANGUAGES C)
+
+enable_testing()
```
maxammann, line 5: Do we use this?

puts/harness/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+
+set(PUTS "")
+set(PUT_LIBS "")
+
+string(REPLACE "," ";" VENDORS "${VENDORS}")
```
maxammann, line 10: What is a vendor?

puts/harness/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+# create Rust init file
+set(RUST_INIT_FILE ${CMAKE_BINARY_DIR}/init.rs)
+set(RUST_INIT_DECL "")
+set(RUST_INIT_CALL "")
+foreach(put_id IN LISTS PUTS)
+    string(APPEND RUST_INIT_DECL "    fn register_${put_id}(callback: FnRegister);\n")
+    string(APPEND RUST_INIT_CALL "        register_${put_id}(callback);\n")
+endforeach()
+
+file(WRITE ${RUST_INIT_FILE} "use crate::FnRegister;\n\nextern {\n${RUST_INIT_DECL}}\n\npub fn register(callback: FnRegister) {\n    unsafe {\n${RUST_INIT_CALL}    }\n}\n")
```
maxammann, line 59: Note to myself: Looks interesting, might need to go back to this.

puts/cmake/CheckCCompilerCanPrintDeps.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+int main()
+{
+    printf("%s\n", "source including system header files");
```
maxammann, line 6:
```suggestion
    printf("source including system header files\n");
```

puts/harness/include/tlspuffin/put.h
```diff
@@ -0,0 +1,155 @@
...
+    const char *version;         // tlspuffin v0.1.0 (commit f80370)
+} C_HARNESS_TYPE;
+
+typedef struct C_LIBRARY_TYPE
+{
+    const char *vendor_name;     // openssl
+    const char *vendor_version;  // OpenSSL 1.1.1j 16 Feb 2021
+
+    const char *config_name;     // openssl111j
+    const char *config_hash;     // da39e788eca8dabb
```
maxammann, line 73: What is a config hash? Do we need to store this field or can we derive from config_name?

puts/harness/include/tlspuffin/put.h
```diff
@@ -0,0 +1,155 @@
...
+
+typedef struct
+{
+    uint8_t name;
+    AGENT_TYPE type;
+    TLS_VERSION tls_version;
+    bool client_authentication;
+    bool server_authentication;
+    const PEM *cert;
+    const PEM *pkey;
```
maxammann, line 55: This might be problematic as this binds everything a lot to TLS which uses certificates.

maxammann: We can leave dynamic data exchange between the fuzzing runtime and the PUT for later.

puts/harness/include/tlspuffin/put.h
```diff
@@ -0,0 +1,155 @@
...
+
+    const PEM *const *const store;
+    const size_t store_length;
```
maxammann, line 58: Which store is this?

puts/harness/src/lib.rs
```diff
@@ -0,0 +1,113 @@
+mod bindings {
+    #![allow(non_upper_case_globals)]
+    #![allow(non_camel_case_types)]
+    #![allow(non_snake_case)]
+    #![allow(improper_ctypes)]
+    #![allow(dead_code)]
+    #![allow(clippy::all)]
```
maxammann, line 7:
```suggestion
    #![allow(warnings, unused)]
```

puts/harness/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+    debug: Some(c_log_debug),
+    trace: Some(c_log_trace),
+    make_result: Some(make_result),
+};
+
+pub unsafe fn to_string(ptr: *const c_char) -> String {
+    use std::ffi::CStr;
+
+    if ptr.is_null() {
+        return "".to_owned();
```
maxammann, line 52:
```suggestion
        return "NULL".to_owned();
```

puts/harness/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+            kind: CErrorKind::Fatal,
+            reason,
+        }),
+        _ => Err(CError {
+            kind: CErrorKind::Error,
+            reason,
+        }),
+    });
+
+    Box::into_raw(result) as *mut _
```
maxammann, line 81: Is this guaranteed to be freed?

maxammann: Maybe we should add a note to the `put.h` file

maxammann (commented):
Naming question: Are we happy with PUT? After reading more it seems like SUT (System under Test) is more common.

I also kind of like "target" but that is too generic.

maxammann:
Continuing review; Accidentally clicked done.

puts/harness/src/lib.c
```diff
@@ -0,0 +1,16 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+void _log(void (*logger)(const char *), const char *format, ...)
+{
+    char *message = NULL;
+    va_list args;
+
+    va_start(args, format);
+    vasprintf(&message, format, args);
```
maxammann, line 11: Can this allocation here fail?

puts/vendor/configs/openssl.toml
```diff
@@ -0,0 +1,52 @@
+[preset]
```
maxammann, line 1: This is really cool :)

puts/vendor/scripts/openssl/CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+  # To be on the safe side, if no prefix is explicitly provided by the caller,
+  # we install the library in an isolated prefix.
+  set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "installation directory" FORCE)
+endif()
+
+option(PARALLEL "Activate parallel build" OFF)
+option(WITH_ASAN "Build with address-sanitizer" OFF)
+option(WITH_SANCOV "Build with sancov" OFF)
+option(WITH_GCOV "Build with instrumentation for gcov coverage" OFF)
+option(WITH_LLVM_COV "Build with instrumentation for llvm coverage" OFF)
```
maxammann, line 29: Wher would these options be set?

shell.nix
```diff
@@ -22,9 +22,6 @@ pkgs.llvmPackages_11.stdenv.mkDerivation {
     # openssh
     pkgs.openssl_1_1
 
-    # Old openssl
-    pkgs.xorg.makedepend
```
maxammann, line 26: Again: Not sure if it works without :D Is it maybe preinstalled on some distros and others it is not?

tlspuffin/Cargo.toml
```diff
@@ -258,6 +260,8 @@ wolfssl = { path = "../crates/wolfssl", optional = true }
 wolfssl-sys = { path = "../crates/wolfssl-sys", features = [], optional = true }
 foreign-types = { version = "0.5.0", optional = true }
 
+tls-harness = { path = "../puts/harness", package="tlspuffin-harness-sys", features = ["tls"], optional = true }
```
maxammann, line 263: Note to myself: Is this generic enough to support e.g. SSH? Why is there "tls" in features?

tlspuffin/src/openssl/mod.rs
```diff
@@ -77,12 +71,47 @@ pub fn new_openssl_factory() -> Box<dyn Factory<TLSProtocolBehavior>> {
...
+                "openssl111j"
+            } else if cfg!(feature = "openssl111u") {
+                "openssl111u"
+            } else if cfg!(feature = "openssl312") {
+                "openssl312"
+            } else if cfg!(feature = "libressl") {
+                "libressl333"
+            } else {
+                "unknown"
+            };
```
maxammann, line 99: I think it would be nice to get the actual openssl version here.

I think the previous implementation actually called the openssl library.

tlspuffin/src/put_registry.rs
```diff
@@ -1,33 +1,46 @@
...
-        } else {
-             crate::tcp::new_tcp_factory
+pub fn tls_registry() -> PutRegistry<TLSProtocolBehavior> {
+    #[cfg(feature = "cputs")]
+    extern "C" fn callback(put: *const C_PUT_TYPE) {
+        if put.is_null() {
+            return;
         }
+
+        println!("registering C put")
```
maxammann, line 18: Is this a debug statement?

maxammann: I guess it is not, but should we maybe use a logger here?

tlspuffin/src/wolfssl/mod.rs
```diff
@@ -71,12 +72,45 @@ pub fn new_wolfssl_factory() -> Box<dyn Factory<TLSProtocolBehavior>> {
...
+                "wolfssl 5.2.0-stable"
+            } else if cfg!(feature = "vendored-wolfssl510") {
+                "wolfssl 5.1.0-stable"
+            } else if cfg!(feature = "vendored-wolfssl430") {
+                "wolfssl 4.3.0-stable"
+            } else if cfg!(feature = "vendored-master") {
+                "master"
+            } else {
+                "unknown"
+            };
```
maxammann, line 98: Same here, I think it would be nice to get the real version that is executed. This could also be nice for verifying a build.

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
...
+NAME="mk_vendor"
+VERSION="0.1"
+
+USAGE="Usage:
+    ${NAME} make <configuration> [vendor-options]
+    ${NAME} locate <configuration> [vendor-options]
+    ${NAME} -h | --help
+    ${NAME} --version
+
+build and interact with vendor libraries
```
maxammann, line 14: This needs to be extended to explain in more detail what this tool is used for.

I guess this is expected to be executed from build.rs files, right?

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
...
+            return
+        fi
+    fi
+
+    if command -v perl > /dev/null 2>&1; then
+        if perl -MCwd -le 'print Cwd::abs_path shift' "$1" > /dev/null 2>&1; then
+            perl -MCwd -le 'print Cwd::abs_path shift' "$1" 2> /dev/null
+            return
+        fi
+    fi
```
maxammann, line 113: For what do we need perl? Can we replace it with e.g. python?

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
...
+
+Display the path to the install prefix that would be created by passing the same
+vendor configuration to '${NAME} make'.
+
+Vendor Options:
+    -n,--name=STRING       override the configuration name
+    -f,--fetch=STRING      override the fetch method used to retrieve the sources
+    -p,--patch=STRING,...  override the patch sets applied to the sources
+    -b,--build=STRING,...  override the build scripts configuration
+    --options=STRING,...   override the build options
```
maxammann, line 77: Can you give examples of which kind of values are possible here?

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
+#!/usr/bin/env bash
```
maxammann, line 1: I guess this is the most opinionated comment of this review: Can we replace this with a Rust script? e.g. https://rust-script.org/

If Rust is too clunky would maybe Python work?

I'm having a really hard time understanding this script even though I'm very used all tools used in this script.

maxammann: This script also has some dependencies like `toml` which kind of makes it harder to use compared to something "batteries included".

I can also support in translating this.

tools/mk_vendor
```diff
@@ -0,0 +1,854 @@
...
+    IFS=',' read -r -a array <<< "${value}"
+
+    # remove quotes from each array element
+    IFS=$'\n' read -r -d '' -a array < <((printf '%s\n' "${array[@]}" | unquote) && printf '\0')
+
+    value=$(IFS="${separator}"; printf '%s\n' "${array[*]}")
+
+    printf "%s\n" "${value}"
+    return 0
+}
```
maxammann, line 602: This seems very manual and easier to do in most other languages.

tools/mk_vendor
maxammann: Note: did not yet fully review this.

puffin/src/protocol.rs
michaelmera: Yes, it was. We mostly referred directly to TLS_PUT_REGISTRY and SSH_PUT_REGISTRY.

I chose to not bind the protocol to a specific registry, because we will be registering the C PUTs dynamically and I would like to stay flexible on which registry is used where.

puts/cmake/CheckCCompilerCanPrintDeps.cmake
michaelmera: It is a helper cmake script to check that the compiler supports the `-M` option.

I have taken note to add some comments to the cmake files in a future PR. I'd rather wait for the entire build chain to be in place, because things might move around a bit. Tell me if you absolutely want it in this PR.

puts/harness/CMakeLists.txt
michaelmera: Not sure what you are talking about, so I'll answer for all lines 😄

- We set a consistent minimal cmake version because it provides a common set of policies on how cmake works and which functions are available.
- We need the C language to get the C toolchain to build the harness and bundle the PUT into a single object file. This allows us to make sure all symbols coming from the vendor library (e.g. OpenSSL) are local, which allows us in turn to link statically several PUTs including the same vendor library into tlspuffin (e.g. OpenSSL 1.1.1k + OpenSSL 1.1.1j)
- We enable testing as it is generally a good practice in cmake builds. It gives access to the `test` target (empty in that case) and allows users to do the generic `make && make test && make install` (or equivalent for other back-ends) without knowing anything about your project.

puts/harness/CMakeLists.txt
michaelmera: That's a very good question. My current naming scheme is:
- the vendor library (or simply vendor) is the library we are testing (e.g. OpenSSL / OpenSSL 1.1.1k). This is why the `mk_vendor` script is called that.
- the harness is the wrapper code that allow us to orchestrate the fuzzing/trace execution on the vendor library (creating server/client, setting agents runtime configuration, ...)
- the PUT is the pair harness + vendor

I'm open to change in that regard. I think here is not the best place to have this discussion, but I would gladly discuss the matter in a specific issue. Or we can setup a meeting. In any case, I would like that we all agree on the naming before I start writing the documentation for tlspuffin, so that I can also rename the Rust traits/structs as necessary and keep the tiny bit of sanity I still maintain 🥺

puts/harness/CMakeLists.txt
michaelmera: This is what will allow us to register the C PUTs in tlspuffin in the future. We generate a tiny Rust file performing just the PUT registration for each of the PUT, and this is included into the `tlspuffin-harness-sys`crate. It can then be called with specific callbacks when creating a `PutRegistry`.

For the record, we first planned to do that using compiler facilities like `__attribute__((constructor))` and made it work. However, I found several mentions that this have been broken in Rust for some versions of clang and that it might not be the way forward that Rust choose to support. Using code generation is future-proof and hopefully more consistent with the way the Rust ecosystem works (e.g. bindgen).

I'm not particularly proud of this specific cmake code though. I intend to change this into a template of sort once I figure out the best way to do that.

puts/harness/CMakeLists.txt
michaelmera: This will come in a future PR. The idea will be that each C PUT harness advertises through cmake which protocol it supports and which vendor library it can wrap (OpenSSL, LibreSSL, ... maybe with version ranges). Then when we get the VENDORS list here, we check which vendor library version it is and select the most appropriate harness. So these variables will not stay hardcoded like this.

puts/harness/include/tlspuffin/put.h
michaelmera: I totally agree with you: this is tls-specific for now and we will need to design a better mechanism in the future. This is intentionally left as future work for now, until everything is working as intended for all the TLS PUTs.

puts/harness/include/tlspuffin/put.h
michaelmera: This is the agent's certificates store. The decision of what to include is left to the Rust part in tlspuffin. In terms of what exists today in the code:
- the server (ALICE) store includes BOB and EVE certificates
- the client (BOB) store includes ALICE and EVE certificates

puts/harness/include/tlspuffin/put.h
michaelmera: The config hash corresponds to the vendor library build configuration. In this PR it is based on the path to the provided library but in the future it will match the configuration selected through `mk_vendor`. The idea is to have a unique identifier for the vendor library, so that we are able to distinguish between different builds.

For now, `mk_vendor` includes the following information in the hash:
- the fetch configuration for the sources (e.g. git url + git ref)
- a hash of the patches applied on these sources
- a hash of the wrapper build script used to build these sources (could be replaced with the version of the tlspuffin workspace once we use a release process enforcing semver)
- the build parameters passed to the wrapper build script (WITH_ASAN, WITH_GCOV, ...)

michaelmera: Forgot to answer the actual question 😅 : We cannot derive it from the name or version.

puts/harness/src/lib.c
michaelmera: Yes. In that case the message will be NULL, which must be handled by the logger callback.

I purposely didn't check for NULL here, because I don't believe a correct decision can be made at this level on how to handle this case: it might be because something went wrong in the vendor library that we cannot allocate memory. In that case, we don't want to fail here in the harness but rather at a higher level in Rust or when giving control back to the vendor library, so that we don't mistake a library bug for a harness bug.

puts/harness/src/lib.rs
michaelmera: I think I want to keep the specific whitelist in that case. Because we have full control on the code from which the bindings are generated, I would like to notice if some new warnings are generated.

puts/harness/src/lib.rs
michaelmera: I want to keep the empty string here, because it makes the C code much simpler in many cases when passing strings to Rust. If you are uncomfortable about this, we can also return a `Result` from `to_string` and let the caller decide what to do.

puts/harness/src/lib.rs
michaelmera: The purpose of this function is actually to make sure that we free the memory corresponding to the result. The only purpose of the tuple `(result_code, description)` is to be passed back to Rust, and the Rust code has no choice but to take ownership back if it wants to read it (and thus it will be freed at some point).

If it was allocated in C, it would be difficult to guarantee that this behaves well in every PUT harness implementation. In particular, note that Rust duplicates the `description` string by calling `to_string` so that all memory pointed by the returned pointer is guaranteed to have been allocated through the Rust allocator.

The only case where memory can be lost is when the C harness don't send the resulting pointer back to Rust, but there is not much we can do about that. In the future, I want to discuss all that in a documentation about how to write C PUTs. The specifics in `put.h` are still subject to changes, so I want to wait before making that kind of statement there.

puts/vendor/configs/openssl.toml
michaelmera: Glad you like it 😄. Hopefully in the future adding support for new configurations will be as easy as adding a few lines here 🥲

puts/vendor/scripts/openssl/CMakeLists.txt
michaelmera: This script is the interface between `mk_vendor` and the actual native build system for the vendor library (e.g. configure + make in the case of OpenSSL). The options themselves are defined in `puts/vendor/configs/openssl.toml` and set through the more convenient interface provided by `mk_vendor`.

So you would do:
```sh
$ mk_vendor make openssl:openssl111j --options=gcov # -> pass WITH_GCOV=ON to cmake
$ mk_vendor make openssl:openssl111j --options=sancov,asan # -> pass WITH_SANCOV=ON + WITH_ASAN=ON to cmake
```

Does that answer your question?

tlspuffin/Cargo.toml
michaelmera: The `tlspuffin-harness-sys` crate will handle the creation of PUTs regardless of protocols. The `tls` feature is here to limit the PUTs that will be linked into `tlspuffin`: We only want the PUTs that implement the TLS protocol. Respectively, in the future, `sshpuffin` will depend on this crate with the `ssh` feature.

Note that we plan to support more fine grain control over which PUTs are included, through environment variables:

- include all PUTs for  vendor libs in $VENDOR_DIR
```sh
    $ [VENDOR_DIR=...] cargo build -p tlspuffin
```

- filter PUTs included in build, by name:
```sh
    $ [VENDOR_DIR=...] [WITH_PUT=openssl111k,wolfssl430-asan,...] cargo build -p tlspuffin
```

Unfortunately, as cargo don't support more complex features, this is the only way I have found to provide this level of granularity, while providing sane defaults.

tlspuffin/src/openssl/mod.rs
michaelmera: Yes, but the change was done on purpose.

Unlike what we now have in Rust, several C PUTs can be present at the same time in tlspuffin. In next PRs, the plan is to introduce a way to select the PUT through the CLI. We will need a shortname to refer to a specific PUT and that is the purpose of this variable.

Note that the features will disappear when Rust PUTs are removed and we plan to use the name given by the user to `mk_vendor` instead. So there will be a single name that the user has chosen to refer to a specific PUT.

tlspuffin/src/put_registry.rs
michaelmera: Right now, it is a placeholder until we include the first real C PUT. I decided to use `println!` so that I'm sure to see this callback being called, regardless of the log level. This only appears when the `cputs` feature is set anyway, so it should not be a disturbance for the current applications.

tlspuffin/src/wolfssl/mod.rs
michaelmera: The shortname is not a substitute for the version. The full version is also provided below, and I adapted the CLI to log both. Here is an example output:
```
2024-03-28T11:06:50+01:00       Put Versions:
2024-03-28T11:06:50+01:00       (Rust) rust-tcp:
2024-03-28T11:06:50+01:00           harness: TCP_______ (tlspuffin 0.1.0+git.1ec568)
2024-03-28T11:06:50+01:00       (Rust) rust-openssl:
2024-03-28T11:06:50+01:00           harness: OPENSSL111 (tlspuffin 0.1.0+git.1ec568)
2024-03-28T11:06:50+01:00           library: openssl (openssl111k / OpenSSL 1.1.1k  25 Mar 2021)
```

README.md
```diff
@@ -98,9 +98,6 @@ coverage, expressiveness of executable protocol traces and stable and extensible
 * clang
 * graphviz
 
-OpenSSL 1.0:
-* makedepend from `xutils-dev package
```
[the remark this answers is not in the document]
michaelmera: Nope. I removed the dependency in #300 (on which this PR is stacked).

I had several instances of makedepend not finding the correct dependencies. In the best case, the build failed but sometimes it caused subtle problems with part of the system OpenSSL being included in our build. Since makedepend was a pain to setup on some platforms, I chose to patch it away definitely.

You can see `puts/vendor/scripts/openssl/CMakeLists.txt` for the changes: I patch the OpenSSL build system and use the compiler to extract dependencies instead. This is the way it is done in more recent releases of OpenSSL and any modern compiler we plan to support should be able to do that.

puffin/src/trace.rs
michaelmera: We don't store a reference anymore but the actual `PutRegistry` (cloned), which allows us to drop the static lifetime. This is made possible by other changes I made to the way the `PutRegistry` is constructed, in order to support the registration of C PUTs.

tools/mk_vendor
michaelmera: It is currently used in the build.rs files of OpenSSL.

However, the longer term idea is to decouple the vendor library build from cargo: the user will first build the vendors he wants (with or without the help of `mk_vendor`) and then call cargo to build tlspuffin.

So, first build whatever you want:
```sh
$ mk_vendor make openssl:openssl111k --name=put1 --options=sancov,asan
$ mk_vendor make openssl:openssl101f --name=put2 --options=sancov,asan --patch=fix-CVE-XYZ
```

Then run:
```sh
$ WITH_PUT=put1,put2 cargo run -p tlspuffin --bin tlspuffin -- ...
```

I agree that this will need a lot more documentation, but I'm waiting for the entire new workflow to be ready before committing to extensive documentation. Also, this might still be subject to changes in the near future.

tools/mk_vendor
michaelmera: It is used as one of the attempts to properly normalize paths in the script. There is no clean way to do it in bash.

It's not a strict dependency, if not present, we use something else, but it is likely that Perl is installed and it is also needed for building OpenSSL.

tools/mk_vendor
michaelmera: I agree with you, this should be rewritten at some point. I second using Rust, so that we don't need to introduce an additional dependency just for this. For now I would keep this so that we can move on and continue the re-architecture of the build, but it's in my todo-list to rewrite it. If you'd like to give it a try, feel free to submit a PR anytime once this one is merged.

For the record, I didn't expect this script to reach this size as it was supposed to be a proof of concept for building vendor libraries independently from cargo. It grew progressively as we noticed we needed more features and we added configuration files, additional commands, default options, etc. In that respect, it has served its purpose: I have been able to quickly change the interface and internals several times until it started to stabilize.

tools/mk_vendor
michaelmera: These options match the structure of the configuration in `openssl.toml`. Each allows you to override the default value set in the presets.

For example: `mk_vendor make openssl:openssl111k --fetch=openssl111j` will override the sources fetch to use the `fetch.openssl111j` value defined in the configuration file instead of the one defined in the selected preset (`fetch.openssl111k`).

In truth, most of the time you won't need to use these options but use the defined presets instead. So I don't want to write too much about them in the help text. Maybe a better approach would be to guide the user through an interactive CLI to help her define a custom preset. This would allow us to provide the list of fetch/build/... values already present in the specific configuration file that she wants to extend.

tools/mk_vendor
michaelmera: Welcome to bash 😄

LCBH:
> Naming question: Are we happy with PUT? After reading more it seems like SUT (System under Test) is more common.
> 
> I also kind of like "target" but that is too generic.

I don't really care, and be happy to change in another PR if you consider this would be better.

tlspuffin/Cargo.toml
LCBH: > Note to myself: Is this generic enough to support e.g. SSH? Why is there "tls" in features?

Beware: this is the Cargo.toml in `tlspuffin` so it is dedicated to tls. Currently, the script and build arch. is `ProtocolBehavior` specific but we could eventually factor out the parts that are not specific in a dedicated crate.

tools/mk_vendor
maxammann: Got it! I think it would be great to have an even more manual process, where you only produce a static Rust library from tlspuffin and then call the linker on your own.

That way it would also possible to distribute binaries and add tooling around it that follow the classical approach of instrumenting/fuzzing.

In AFL++ for instance you have `afl-cc` that you can use to compile your program. After compilation it is no longer a program but a fuzzing target that can be executed. Much like `clang -fsanitize=fuzzer`. If we split all of the steps sufficiently, we can have a `tlspuffin-cc` that you can set using `CC=tlspuffin-cc` and then compile e.g. a patched OpenSSL.

puts/harness/include/tlspuffin/put.h
maxammann: Can you add a todo note there?

michaelmera: I just pushed an additional TODO in this file to mention the need to exclude the TLS-specific constructs.

tools/mk_vendor
michaelmera: I opened #308 to discuss that further. In any case, I agree that we might want to provide smaller build steps at some point, but I think the proposed design matches our current needs without excluding that possibility.

puts/harness/CMakeLists.txt
michaelmera: I thought about this and I agree with you. In this context it makes more sense to talk about a "library" that we want to wrap into our harness code. I modified it here and in the calling context of this CMakeLists.txt.

PR #265, opened 2023-06-09
Add complete trace with 3rd step for CVE-2022-39173

maxammann:

PR #306, opened 2024-03-22
fix CI deadlock when testing known vulnerabilities

michaelmera:
After some investigation on why CI would block on #304 and #290, it is the same problem encountered in #265: asan is randomly deadlocking on us. It seems to only happen every time `test_wolfssl_openssl_test_seed_cve_2022_38153` is not run last. Running rustc with `--test-threads 1` apparently solves the problem but is not convenient for the current CI scripts.

This PR introduces a timeout argument for the tests on known vulnerabilities and retries several times if we fail to observe the crash. Until the problem is fixed upstream, this seems like the most sensible way to handle the problem without disabling the tests.

See also:
- upstream issue: llvm/llvm-project#63224

### DIFF
--- a/tlspuffin/src/integration_tests/mutations.rs
+++ b/tlspuffin/src/integration_tests/mutations.rs
@@ -1,6 +1,7 @@
 use puffin::{
     agent::AgentName,
     algebra::{dynamic_function::DescribableFunction, Term},
+    execution::{forked_execution, AssertExecution},
     fuzzer::mutations::{
         util::TermConstraints, RemoveAndLiftMutator, RepeatMutator, ReplaceMatchMutator,
         ReplaceReuseMutator,
@@ -18,7 +19,6 @@ use puffin::{
 use crate::{
     put_registry::TLS_PUT_REGISTRY,
     query::TlsQueryMatcher,
-    test_utils::expect_crash,
     tls::{
         fn_impl::{
             fn_client_hello, fn_encrypt12, fn_seq_1, fn_sign_transcript,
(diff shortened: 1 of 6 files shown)


tlspuffin/src/integration_tests/mutations.rs
```diff
@@ -46,133 +46,167 @@ fn test_mutate_seed_cve_2021_3449() {
...
-                if let Action::Input(input) = &step.action {
-                    if input.recipe.name() == fn_encrypt12.name() {
-                        return true;
+                fn check_is_encrypt12(step: &Step<TlsQueryMatcher>) -> bool {
+                    if let Action::Input(input) = &step.action {
+                        if input.recipe.name() == fn_encrypt12.name() {
+                            return true;
+                        }
                     }
+                    false
```
maxammann, line 66: @michaelmera just so that you know, these tests here were an effort during the last weeks of paper writing to find reasons for fuzzing limitations. So they are low quality and likely should go away in the future.

maxammann:
This one is very annoying. I think it is fine to ignore that for now, but it is unlikely this will get fixed ever.

michaelmera:
> This one is very annoying. I think it is fine to ignore that for now, but it is unlikely this will get fixed ever.

Yes, that's unfortunate but we don't have much influence on the priorities of fixes in llvm. I also don't think it would be reasonable to spend the time to investigate the internals of asan further in order to develop a fix.

Anyway, I think it is a good practice to set a timeout and retry executions where we expect a crash. So I guess the time spent here was not in vain.

Issue #307, opened 2024-03-29
Status of BoringSSL


aeyno:
There are still some things to do to fully support BoringSSL (#290) :
- Debug builds with ASAN don't work without `lto=true`
- The determinism test doesn't pass with multithreaded `cargo test` but the determinism is working when tested in a single threaded manner and also while fuzzing (thanks to the fact that libafl uses forks)
- The security claims are not yet implemented
- Some seeds are not working due to the way BoringSSL splits its handshake packets

maxammann:
> * Some seeds are not working due to the way BoringSSL splits its handshake packets

Here are some notes on how we can make boringssl behave like other engines.

https://github.com/tlspuffin/tlspuffin/pull/290#discussion_r1541392572

maxammann:
Ideally, we can remove the additional seeds and cfg() options related to boringssl [here](https://github.com/tlspuffin/tlspuffin/pull/290#discussion_r1541337867).

PR #309, opened 2024-04-05
fix OpenSSL deterministic RNG

michaelmera:
Due to previous changes introduced in #290, we stopped reseeding the RNG (short seed was silently rejected). This has gone unnoticed for some time and finally broke to light in CI for #280.

This PR:
- ensures that OpenSSL's RNG is reseeded when requested by providing a seed long enough.
- bypasses libc's `RAND_seed` function when reseeding from Rust code.
- splits the `set_openssl_deterministic` function into `rng_set` and `rng_reseed` to match higher level constructs.
- avoids a race condition when reseeding the deterministic RNG in multithreaded environments (e.g. `cargo test`).
- adds some more tests to the deterministic RNG to prevent future regression.

### DIFF
--- a/crates/openssl-src-111/src/deterministic_rand.c
+++ b/crates/openssl-src-111/src/deterministic_rand.c
@@ -3,8 +3,23 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+#ifndef thread_local
+// since C11 the standard include _Thread_local
+#if __STDC_VERSION__ >= 201112 && !defined __STDC_NO_THREADS__
+#define thread_local _Thread_local
+
+// note that __GNUC__ covers clang and ICC
+#elif defined __GNUC__ || defined __SUNPRO_C || defined __xlC__
+#define thread_local __thread
+
+#else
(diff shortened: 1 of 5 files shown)


crates/openssl-src-111/src/deterministic_rand.c
```diff
@@ -1,25 +1,37 @@
...
 
-static uint64_t seed = 42;
+#ifndef thread_local
+// since C11 the standard include _Thread_local
+#if __STDC_VERSION__ >= 201112 && !defined __STDC_NO_THREADS__
+#define thread_local _Thread_local
+
+// note that __GNUC__ covers clang and ICC
+#elif defined __GNUC__ || defined __SUNPRO_C || defined __xlC__
+#define thread_local __thread
```
LCBH, line 13: Could we use that to solve the second item of #307?! @aeyno

aeyno: Done, see #310.

PR #310, opened 2024-04-06
Fixing BoringSSL randomness in multithread execution

aeyno:
This PR fixes the randomness issue (mentionned in #307) when executing BoringSSL in multithread using thread_local variables, like what has been done in #309. It also enables the tests associated with the BoringSSL randomness.

### DIFF
--- a/crates/boringssl-src/patches/reset_drbg.patch
+++ b/crates/boringssl-src/patches/reset_drbg.patch
@@ -1,8 +1,8 @@
 diff --git a/crypto/rand_extra/deterministic.c b/crypto/rand_extra/deterministic.c
-index d1d582b07..a6e1d092d 100644
+index d1d582b07..a0e87f8d5 100644
 --- a/crypto/rand_extra/deterministic.c
 +++ b/crypto/rand_extra/deterministic.c
-@@ -24,16 +24,46 @@
+@@ -24,23 +24,50 @@
  
  #include "../internal.h"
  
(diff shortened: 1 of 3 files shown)


maxammann (approved):

crates/boringssl-src/patches/reset_drbg.patch
```diff
@@ -1,28 +1,43 @@
...
+ 
+ // g_num_calls is the number of calls to |CRYPTO_sysrand| that have occurred.
+ //
+ // This is intentionally not thread-safe. If the fuzzer mode is ever used in a
+ // multi-threaded program, replace this with a thread-local. (A mutex would not
+ // be deterministic.)
+-static uint64_t g_num_calls = 0;
+-static CRYPTO_MUTEX g_num_calls_lock = CRYPTO_MUTEX_INIT;
++static thread_local uint64_t g_num_calls = 0;
++static thread_local CRYPTO_MUTEX g_num_calls_lock = CRYPTO_MUTEX_INIT;
```
michaelmera, line 31: Any reason to keep this mutex at all now that the guarded variable is thread local?

michaelmera (commented):
Seems good. You might want to remove the remaining mutex.

crates/boringssl-src/patches/reset_drbg.patch
aeyno: You're right. I removed it

michaelmera (approved):

PR #313, opened 2024-04-29
Adding flights of packets to the knowledge and ability to decrypt a whole flight of packets

aeyno:
This PR allows storing a flight of packet, eg. all packets that were sent by a PUT between two steps, in the knowledge. This flight could then be queried and used in a trace as a `Vec<Message>`. This PR also adds the ability to send a Flight of packets in a trace step, which allows to send multiple packets at once.

This is needed to make the traces less dependent of implementation choices and to be resilient if some implementations choose to fragment their packets in different ways.

The first use case is to create traces that work both for OpenSSL/WolfSSL/LibreSSL and BoringSSL, the latter fragmenting his handshake messages differently than the formers.


puffin/src/protocol.rs
```diff
@@ -1,14 +1,97 @@
...
+
+impl<O: OpaqueProtocolMessage> Codec for OpaqueMessageFlight<O> {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        for msg in &self.messages {
+            msg.encode(bytes);
+        }
+    }
+
+    fn read(_reader: &mut Reader) -> Option<Self> {
+        None
```
LCBH, line 83: Please implement a proper reader here, simply reading the buffer until exhaustion.

puffin/src/trace.rs
```diff
@@ -614,19 +632,35 @@ impl<M: Matcher> InputAction<M> {
...
             .as_ref()
             .downcast_ref::<PB::OpaqueProtocolMessage>()
         {
             opaque_message.debug("Input opaque message");
-            ctx.add_to_inbound(step.agent, opaque_message)?;
+            ctx.add_to_inbound(step.agent, &opaque_message.clone().into())?;
         } else {
             return Err(FnError::Unknown(String::from(
-                "Recipe is not a `ProtocolMessage`, `OpaqueProtocolMessage`!",
+                "Recipe is not a `ProtocolMessage`, `OpaqueProtocolMessage`, `MessageFlight`!",
```
LCBH, line 663: ,or OpaqueMessageFlight

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -50,34 +50,56 @@ pub fn fn_append_transcript(
     Ok(new_transcript)
 }
 
-pub fn fn_decrypt_handshake(
-    application_data: &Message,
+pub fn fn_new_flight() -> Result<MessageFlight<Message, OpaqueMessage>, FnError> {
```
LCBH, line 53: Don't we need those two last function symbols for OpaqueMessageFlight as well?

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -50,34 +50,56 @@ pub fn fn_append_transcript(
...
+                client,
+                &sequence_number,
+            )?;
+
+            decrypted_flight.append(&mut decrypted_msg);
+            sequence_number += 1;
+        }
+    }
+
+    println!("Decrypting handshake");
```
LCBH, line 97: Use logging (I'd say trace!()) instead

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -50,34 +50,56 @@ pub fn fn_append_transcript(
...
+    for msg in &decrypted_flight {
+        println!("DECRYPTED: {:?}", msg);
```
LCBH, line 99: same here

LCBH: actually remove that for loop

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -197,6 +219,46 @@ pub fn fn_psk(some: &Vec<u8>) -> Result<Option<Vec<u8>>, FnError> {
...
+                client,
+                &sequence_number,
+            )?;
+
+            decrypted_flight.push(decrypted_msg);
+            sequence_number += 1;
+        }
+    }
+
+    for msg in &decrypted_flight {
```
LCBH, line 255: remove

tlspuffin/src/tls/seeds.rs
```diff
@@ -266,6 +267,49 @@ pub fn seed_successful(client: AgentName, server: AgentName) -> Trace<TlsQueryMa
...
+                        (client, 1)/MessageFlight<Message,OpaqueMessage>
+                    },
+                }),
+            },
+        ],
+    }
+}
+
 /// Seed which triggers a MITM attack. It changes the cipher suite. This should fail.
 pub fn seed_successful_mitm(client: AgentName, server: AgentName) -> Trace<TlsQueryMatcher> {
```
LCBH, line 314: This is subsumed by the previous one, no?

tlspuffin/src/tls/seeds.rs
```diff
@@ -2034,18 +1717,22 @@ pub fn seed_session_resumption_dhe_full(
         )
     };
 
-    let resumption_encrypted_extensions = term! {
-        fn_decrypt_handshake(
-            ((server, 0)[Some(TlsQueryMatcher::ApplicationData)]), // Encrypted Extensions
+    let resumption_decrypted_handshake = term! {
```
LCBH, line 1720: rename this

tlspuffin/src/tls/seeds.rs
```diff
@@ -2160,16 +1839,15 @@ pub fn create_corpus() -> Vec<(Trace<TlsQueryMatcher>, &'static str)> {
     corpus!(
         // Full Handshakes
         seed_successful: cfg(feature = "tls13"),
+        seed_successful_with_flights: cfg(feature = "tls13"),
```
LCBH, line 1842: do not have add redundant traces here, so I guess remove seed_successful.

LCBH: Eventually, rename `seed_successful_with_flights` into `seed_successful`

LCBH (changes requested):

tlspuffin/src/tls/fn_utils.rs
aeyno: yes I already removed that in another commit in the new-queries branch, I forgot some debug prints

tlspuffin/src/tls/seeds.rs
aeyno: seed_successful_mitm modifies the client hello and forward the other messages of the handshake, while seed_successful just forward a complete handshake.
Is it normal that this trace is called seed_**successful**_mitm knowing that this trace should fail (cf. the comment above the trace) ?

puffin/src/protocol.rs
aeyno: Done with the new traits `ProtocolMessageFlight` and `OpaqueProtocolMessageFlight`

tlspuffin/src/tls/seeds.rs
LCBH: OK

tlspuffin/src/tls/seeds.rs
aeyno: I removed `seed_successful` and renamed `seed_successful_with_flights` into `seed_successful`.  Since `seed_successful` was used in `seed_successful_with_ccs`, I changed `seed_successful_with_ccs` to be a stand-alone trace.

PR #316, opened 2024-05-13
Minor docs tweaks

maxammann:


michaelmera (approved):
Seems good to me! We will probably need a second pass later to avoid redundant information once we have a reference guide.

Note 1: I fixed a few minor typos directly in the PR.
Note 2: I will write an component to include code fragments directly from Rust files, so that the sources and the doc don't go out of sync. But that can wait a futtur PR.

PR #328, opened 2024-09-06
Upgrade MSRV to 1.70 and libafl to 11.0

michaelmera:
This PR will let us build the project with the full range of Rust versions starting from 1.70 to the current nightly. This was previously broken because of a size assertion on [`TypeId`](https://doc.rust-lang.org/std/any/struct.TypeId.html) in `libafl < 11.0`

This will let us upgrade the project's tooling and dependencies. For example,`rust-analyzer` was broken on the project starting from Rust 1.70 (see https://github.com/rust-lang/rust-analyzer/issues/17910).

Most notable changes:
- explicitly set the MSRV (1.70)
- explicitly test the MSRV in CI
- cleanup unused project dependencies


tlspuffin/src/integration_tests/mod.rs
```diff
@@ -42,6 +42,6 @@ fn version_test() {
     assert!(version.expect("missing version string").contains("5.3.0"));
     #[cfg(feature = "wolfssl540")]
     assert!(version.expect("missing version string").contains("5.4.0"));
-    #[cfg(feature = "wolfssl550")]
```
LCBH, line 45: Why removing wolfssl550?

rust-toolchain.toml
```diff
@@ -1,3 +1,11 @@
 [toolchain]
-channel = "1.68.2"
-targets = [ "x86_64-unknown-linux-gnu" ]
+channel = "nightly-2024-09-05"
```
LCBH, line 2: Why nightly?

LCBH (changes requested):

.github/workflows/run-checks.yml
```diff
@@ -56,3 +56,19 @@ jobs:
         run: just check-crate wolfssl-sys wolfssl540,fix-CVE-2022-39173,fix-CVE-2022-42905
       - name: check `boringssl-sys`
         run: just check-crate boringssl-sys boringssl202403
+
+  msrv:
+    name: msrv
+    runs-on: ubuntu-22.04
```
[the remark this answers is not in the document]
michaelmera: Good point. For now, all CI runs in ubuntu 22.0 so I'll do a different PR to upgrade it to ubuntu 24.04 and also investigate the problem about duplicated CI runs that you reported.

tlspuffin/src/integration_tests/mod.rs
michaelmera: We actually stop at 5.4.0, so this was triggering a warning flagging this as a non-existing feature. I think I'll write a more generic version of this test once we merge the dynamic PUT registration process.

rust-toolchain.toml
michaelmera: This ensures that we get access to all necessary features of the Rust tools. Unfortunately, many useful features are only accessible in nightly (e.g. for `rustfmt` and `rust-analyzer`).

Note that this is only the default toolchain, you can override it on the CLI with the `+toolchain` pattern or with the `RUSTUP_TOOLCHAIN` environment variable.

LCBH (approved):
LGTM

PR #330, opened 2024-09-10
Fix integration tests

michaelmera:
To execute traces that can crash or hang we currently provide a helper to run a trace in a sub-process until the execution completes or a timeout duration is reached. This PR addresses the failing/deadlocking problems encountered in CI due to the unreliability of the current implementation.

Following is a summary of the changes included in this PR.

### 1. implement a reliable timeout for forked processes

The current timeout mechanism relied on multi-threading in the child worker process. This worked well for successful executions and crashes but in rare cases the worker process hanged in a way that prevented the timeout signal to trigger, leaving the worker process hanging.

This PR now spawns a watchdog process alongside the worker process, responsible for signaling the timeout to the parent process. After some investigation this appeared as the most reliable solution in face of possible multi-threading in the parent process.

### 2. ensure that integration tests are run single-threaded

Even with the watcher process, there is still no guarantee that the rust code running in the child process will behave as expected. Thus, we leverage cargo `tests` directory convention to ensure that the integration tests are actually detected as such. `cargo` will take care of running such tests sequentially in a single-threaded environment.

### 3. better separation of concerns for the `Trace` execution process

To implement the forked execution, this PR also introduces a `TraceRunner` trait which serves as the new entry point for executing a `Trace`.


tlspuffin/src/test_utils.rs
```diff
@@ -1,69 +1,251 @@
...
+        }
+    }
+
+    pub fn wolfssl_server(port: u16, version: TLSVersion) -> ParametersGuard {
+        let (_key, _cert, temp_dir) = gen_certificate();
+
+        let port_string = port.to_string();
+        let mut args = vec!["-p", &port_string, "-x", "-d", "-i"];
+        let prog = "./examples/server/server";
+        let cwd = "/home/max/projects/wolfssl";
```
aeyno, line 162: Maybe remove this absolute path

aeyno (commented):
LGTM

tlspuffin/src/openssl/mod.rs
```diff
@@ -167,25 +166,30 @@ impl Put<TLSProtocolBehavior> for OpenSSL {
         result
     }
 
-    fn reset(&mut self, _agent_name: AgentName) -> Result<(), Error> {
+    fn reset(&mut self, new_name: AgentName) -> Result<(), Error> {
+        self.config.descriptor.name = new_name;
+
+        #[cfg(feature = "claims")]
+        self.deregister_claimer();
```
LCBH, line 173: Not sure we want to remove al lclains from self!

LCBH (changes requested):

tlspuffin/src/test_utils.rs
michaelmera: We should definitely do something about this. I noticed several occurrences, so I opened #335 to take the time and evaluate what we should do with these in a future PR.

tlspuffin/src/protocol.rs
```diff
@@ -344,7 +344,7 @@ impl ExtractKnowledge<TlsQueryMatcher> for HandshakePayload {
             data: Box::new(self.clone()),
         });
         match &self {
-            HandshakePayload::HelloRequest => {}
+            HandshakePayload::HelloRequest | HandshakePayload::HelloRetryRequest(_) => {}
```
LCBH, line 347: Why did you have to change that?

LCBH (changes requested):

tlspuffin/src/openssl/mod.rs
michaelmera: This only removes the claimer callback from the OpenSSL `ssl` object before performing the stream reset. Because when we do a hard reset the `ssl` object might be dropped in OpenSSL, it's safer to first retrieve and drop the callback allocated in Rust.

LCBH: You're right!

.github/workflows/run-validation.yml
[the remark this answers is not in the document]
LCBH: Documented in #336, closing this.

tlspuffin/src/boringssl/deterministic.rs
```diff
@@ -9,24 +9,16 @@ pub fn reset_rand() {
...
 #[cfg(test)]
 mod tests {
-    use puffin::trace::TraceContext;
-
-    use crate::put_registry::tls_registry;
+    use crate::test_utils::prelude::*;
     use crate::tls::seeds::seed_client_attacker_full;
-    use crate::tls::trace_helper::TraceHelper;
 
-    // TODO: This test only works in a single threaded cargo test execution
```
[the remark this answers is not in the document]
michaelmera: Well, this TODO seems wrong to me: we've had this test in CI and locally for months and I've not seen any problem. @aeyno Any idea why and when this used to fail?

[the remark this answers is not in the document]
aeyno: It was failing before #310, now this should be fixed. I probably forgot to remove the TODO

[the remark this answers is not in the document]
LCBH: Ok, makes sense

tlspuffin/src/protocol.rs
LCBH: This is required to make uni tests pass for BoringSSL and WolfSSL, these tests were not performed before this PR.

LCBH (approved):
LGTM

puffin/src/trace.rs
```diff
@@ -225,6 +225,78 @@ impl<PB: ProtocolBehavior> KnowledgeStore<PB> {
...
+    descriptors: HashMap<AgentName, PutDescriptor>,
+    default: PutDescriptor,
+}
+
+impl<PB: ProtocolBehavior> Spawner<PB> {
+    pub fn new(registry: impl Into<PutRegistry<PB>>) -> Self {
+        let registry = registry.into();
+
+        Self {
+            default: PutDescriptor::from(registry.default().name()),
```
LCBH: Bug was here!

PR #337, opened 2024-09-30
Precomputations

aeyno:
This PR adds precomputations capabilities to puffin.

The idea of precomputations is to be able to compute a costly term only once, add it to the knowledge and query it in the rest of the trace.

- `input_action!` macro, allowing to add precomputations to a trace
	```rust
	input_action! {
		"decrypted_extensions" = term! {
			@extensions // any part of the term extensions could be queried using the "decrypted_extensions" label
		}
	    =>
	    term! {
		// here we query a sub-element of the decrypted_extensions precomputations
        	(!"decrypted_extensions", 0)[Some(TlsQueryMatcher::Handshake(Some(
            	HandshakeType::EncryptedExtensions
        	)))] / Message
	    }
	}
	```

- An extension of the `term` macro syntax, allowing to search for precomputations
    ```rust
    (!"precomputation name", 0)[Matcher] / Optional_rust_type
    ```
- A TLS test trace that use precomputations : `client_attacker_full_precomputations`

Currently, precomputations could not be used for fuzzing, because they can't be mutated and would limit the ability of the fuzzer to find new interesting traces. A following PR will implement a way for the fuzzer to mutate precomputations as well as the ability to query arbitrary knowledges.

### DIFF
--- a/puffin/src/trace.rs
+++ b/puffin/src/trace.rs
@@ -62,14 +62,14 @@ impl<M: Matcher> fmt::Display for Query<M> {
 #[derive(Debug, PartialEq, Eq, Clone, Hash, Deserialize, Serialize)]
 pub enum Source {
     Agent(AgentName),
-    Label(String),
+    Label(Option<String>),
 }
 
 impl fmt::Display for Source {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Self::Agent(x) => write!(f, "agent:{}", x),
-            Self::Label(x) => write!(f, "label:{}", x),
+            Self::Label(x) => write!(f, "label:{:?}", x),
         }
     }
 }
@@ -646,7 +646,7 @@ impl<PT: ProtocolTypes> fmt::Display for OutputAction<PT> {
 #[derive(Serialize, Deserialize, Clone, Debug, Hash)]
 #[serde(bound = "PT: ProtocolTypes")]
 pub struct Precomputation<PT: ProtocolTypes> {
-    pub label: String,
+    pub label: Option<String>,
     pub recipe: Term<PT>,
 }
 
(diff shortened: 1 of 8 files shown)


puffin/src/trace.rs
```diff
@@ -659,6 +702,106 @@ impl<PT: ProtocolTypes> fmt::Display for InputAction<PT> {
...
+/// # use puffin::term;
+/// # use crate::algebra::test_signature::fn_seq_0;
+///
+/// InputAction {
+///     recipe: term!{fn_seq_0()},
+///     precomputations: vec![],
+/// };
+/// InputAction {
+///     recipe: term!{fn_seq_0()},
+///     precomputations: vec![Precomputation{label: "".into(), recipe: term!{fn_seq_0()}}],
```
LCBH, line 742: Don't we rather want an `Option<String>` as a field type for `label`?

puffin/src/trace.rs
```diff
@@ -659,6 +702,106 @@ impl<PT: ProtocolTypes> fmt::Display for InputAction<PT> {
...
+///     precomputations: vec![Precomputation{label: "this_is_a_label".into(), recipe: term!{fn_seq_0()}}],
+/// };
+/// InputAction {
+///     recipe: term!{fn_seq_0()},
+///     precomputations: vec![
+///         Precomputation{label: "this_is_a_label".into(), recipe: term!{fn_seq_0()}},
+///         Precomputation{label: "".into(), recipe: term!{fn_seq_0()}}
+///     ],
+/// };
+/// ```
```
LCBH, line 755: Maybe start the DOC with an actually useful example?

LCBH (changes requested):

puffin/src/trace.rs
aeyno: Done

puffin/src/trace.rs
aeyno: I improved the documentation, I hope now it's more explicit

LCBH (approved):
LGTM

PR #342, opened 2024-10-29
refactor: add function attributes in signature

LCBH:
- Add `FunctionAttributes` to symbols in signature, use it to make is_opaque, is_list, is_get protocol-dependent
 - Add `FunctionAttributes` to FnContainer`
 
 Part of #280


LCBH:
Subsumed by #343

Issue #260, opened 2023-06-08
Revaluate removed debug asserts


maxammann:
https://github.com/tlspuffin/tlspuffin/commit/fdbe01262ace6f4517e13bc1408be857b1749446

LCBH:
See https://github.com/tlspuffin/tlspuffin/pull/343/files#r1930511204

PR #343, opened 2024-10-29
bit mutations/refactor

LCBH:
Subset of #280 with refactoring we want to merge in `master` in the short term.


puffin/src/agent.rs
```diff
@@ -219,8 +220,8 @@ impl<PB: ProtocolBehavior> Agent<PB> {
 }
 
 impl<PB: ProtocolBehavior> Stream<PB> for Agent<PB> {
-    fn add_to_inbound(&mut self, message_flight: &PB::OpaqueProtocolMessageFlight) {
-        self.put.add_to_inbound(message_flight)
+    fn add_to_inbound(&mut self, message: &ConcreteMessage) {
```
LCBH, line 223: Consider making `take_message_from_outbound` outputs a ConcreteMessage too so that:
 -  we can get rid of protocol-dependent boiler-plate code for reading the buffer into a `OpaqueProtocolMessageFlight`
  - and Stream does not depend on `ProtocolBehavior` anymore

puffin/src/algebra/mod.rs
```diff
@@ -779,61 +783,76 @@ mod tests {
         let _string = dynamic_fn(&vec![Box::new(1u8)])
             .unwrap()
             .as_any()
-            .downcast_ref::<u16>()
-            .unwrap();
+            .downcast_ref::<u16>();
+        // .unwrap(); TODO: why it used to work annd ow now?
```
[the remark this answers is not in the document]
LCBH: Resolved by removing `use crate::protocol::AsAny;`

aeyno (approved):
LGTM

puffin/src/algebra/bitstrings.rs
```diff
@@ -0,0 +1,256 @@
...
+                           * replace in this sub-tree */
+        }
+    }
+
+    #[must_use]
+    pub fn init_with_path(path: TermPath) -> Self {
+        let mut e_t = Self::init();
+        e_t.path = path;
+        e_t
+    }
```
michaelmera, line 62: I'm not a fan of these constructors: when I found usage of them in the code, I had to go back to these declarations to check what they were doing. I think mostly the `init` part is confusing, I would have expected some other initialization (like a global cache or something similar) rather than a plain constructor. My suggestions:
- renaming `init` to `empty` would make more sense, or implement and use the `Default` trait for that purpose.
- renaming `init_with_path` to `with_path` or implement the `From` trait for `TermPath` (but I think the intent is not really clear in this case)

puffin/src/algebra/bitstrings.rs
```diff
@@ -0,0 +1,256 @@
...
+            let path = &path[1..];
+            if self.args.len() <= nb {
+                Err(Error::Term(format!(
+                    "[replace_payloads] [get] Should never happen! EvalTree: {self:?}\n, path: {path:?}"
+                )))
+            } else {
+                self.args[nb].get(path)
+            }
+        }
+    }
```
michaelmera, line 79: If you happen to change things in the PR before merge, I'd also rather not have these cascading if-else, as this was difficult to read. Just returning in each `if` part would highlight the main execution path (forwarding the rest of the path to the correct child) and draw these parts as special cases.

puffin/src/algebra/bitstrings.rs
```diff
@@ -0,0 +1,256 @@
...
+        &self,
+        eval_tree: &mut EvalTree,
+        path: TermPath,
+        ctx: &TraceContext<PB>,
+        with_payloads: bool,
+        sibling_has_payloads: bool,
+        type_term: &TypeShape<PT>,
+    ) -> Result<(Box<dyn EvaluatedTerm<PT>>, Vec<PayloadContext<PT>>), Error>
+    where
+        PB: ProtocolBehavior<ProtocolTypes = PT>,
```
michaelmera, line 115: This is the biggest change in this PR and it was a bit difficult to follow. One point that confused me is passing both `eval_tree` and `path` while it seems to me that `path` is always equal to `eval_tree.path`. So I'd rather remove this argument to avoid the confusion.

puffin/src/algebra/bitstrings.rs
```diff
@@ -0,0 +1,256 @@
...
+                if with_payloads && (path.is_empty() || (self.payloads.is_some())) {
+                    if let Some(payload) = &self.payloads {
+                        log::trace!("        / We retrieve evaluation for eval_tree from payload.");
+                        eval_tree.encode = Some(payload.payload_0.clone().into());
+                    } else {
+                        let eval = PB::any_get_encoding(d.as_ref());
+                        log::trace!("        / No payload so we evaluated into: {eval:?}");
+                        eval_tree.encode = Some(eval);
+                    }
+                    if with_payloads && self.payloads.is_some() {
```
michaelmera, line 160: The enclosing `if` already ensures that `with_payload` is `true`. Remove the unnecessary check.

puffin/src/algebra/bitstrings.rs
```diff
@@ -0,0 +1,256 @@
...
+                let mut dynamic_args: Vec<Box<dyn EvaluatedTerm<PT>>> = Vec::new(); // will contain all the arguments on which to call the function symbol
+                                                                                    // implementation
+                let mut all_payloads = vec![]; // will collect all payloads contexts of arguments (except those under opaque
+                                               // function symbols)
+                let mut eval_tree_args = vec![]; // will collect the eval tree of the sub-terms, if `with_payloads`
+                let self_has_payloads_wo_root = self.has_payload_to_replace_wo_root();
+                for (i, ti) in args.iter().enumerate() {
+                    log::trace!("  + Treating argument # {i} from path {path:?}...");
+                    if with_payloads && self.is_opaque() && ti.has_payload_to_replace() {
+                        // Fully evaluate this sub-term and consume the payloads
```
michaelmera, line 186: I'm still a bit confused by this part.
Why do we check if the opaque symbol has payloads? Wouldn't it be beneficial to evaluate it concretely in every case so that we don't keep a part of the tree that will not be useful later? Or is it because it might be symbolic? In that case why not test it instead?

puffin/src/algebra/term.rs
```diff
@@ -26,135 +34,484 @@ pub enum Term<PT: ProtocolTypes> {
...
+}
 
-    pub fn get_type_shape(&self) -> &TypeShape<PT> {
-        match self {
-            Term::Variable(v) => &v.typ,
-            Term::Application(function, _) => &function.shape().return_type,
+pub fn has_payload_to_replace_rec<PT: ProtocolTypes>(term: &Term<PT>, include_root: bool) -> bool {
+    if let (Some(_), true) = (&term.payloads, include_root) {
+        return true;
+    } else {
```
michaelmera, line 318: Remove `else` and just dedent its body to make the execution path clearer.

puffin/src/algebra/term.rs
```diff
@@ -26,135 +34,484 @@ pub enum Term<PT: ProtocolTypes> {
...
+        match &term.term {
+            DYTerm::Variable(_) => {}
+            DYTerm::Application(_, args) => {
+                if !term.is_opaque() {
+                    for t in args {
+                        if has_payload_to_replace_rec(t, true) {
+                            return true;
+                        }
+                    }
+                }
```
michaelmera, line 328: What happens if `is_opaque` is `true` and it contains a payload?

puffin/src/codec.rs
```diff
@@ -195,6 +254,69 @@ pub fn encode_vec_u8<T: Codec>(bytes: &mut Vec<u8>, items: &[T]) {
...
+//     fn encode(&self, bytes: &mut Vec<u8>) {
+//         if let Ok(value) = self {
+//             value.encode(bytes);
+//         }
+//     }
+//
+//     fn read(r: &mut Reader) -> Option<Self> {
+//     // ???
+//     }
+// }
```
michaelmera, line 303: Remove it then ;-)

puffin/src/algebra/bitstrings.rs
LCBH: There are several cases:
 1. If self is opaque AND t_i has payload, then we don't have much choices. Because t is opaque, we need to replace the payloads in t_i right away, before applying the opaque function from self (otherwise, we won't find the payload anymore in the evaluation of self). We deal with this case here. Doing this involves fully evaluating t_i (`evaluate`) and then `try_read_bytes` the resulting bitstring.
2. If self is opaque but t_i has no payload to replace, then we prefer not evaluating and then `try_read_bytes` because the latter might not be complete (it's a `try`!) and because we can proceed the old way: only working with rustls message.
3. If self is not opaque, then we can collect the payloads from t_i and replace them later, possibly even at the very end when we have traversed the full term. Again, we prefer not evaluating and  `try_read_bytes` in this case to minimize potential failures.

puffin/src/algebra/term.rs
LCBH: In that case, we do not have to replace those payloads since they will be replaced when evaluating that sub-term which is opaque.

tlspuffin/Cargo.toml
```diff
@@ -289,6 +289,9 @@ client-authentication-transcript-extraction = []
 
 introspection = ["puffin/introspection"]
 
+# Enable rustls guards rejecting ill-formed messages, those were disabled by default to allow the fuzzer to explore more messages
+enable-guards = []
```
LCBH, line 293: #260 Use that feature flag and re-enable those debug assert.

PR #345, opened 2024-10-31
Differential fuzzing

aeyno:
This PR adds support for differential fuzzing using Puffin.
Currently, only TLS is supported with wolfSSL and OpenSSL CPUTs.


puffin/src/claims.rs
```diff
@@ -94,6 +108,91 @@ impl<C: Claim> ClaimList<C> {
         claim.set_step(None);
         self.claims.push(claim);
     }
+
+    pub fn compare(&self, other: &Self) -> Result<(), Vec<TraceDifference>> {
+        let blacklist = <C::PT as ProtocolTypes>::differential_fuzzing_claims_blacklist();
```
nataelbaffou, line 114: Maybe add some comments to explain the logic behind the processing of the claims, like explained if the conf (dedupping, ordering, ...)

puffin/src/cli.rs
```diff
@@ -80,7 +81,9 @@ where
                 .arg(arg!(-c --show_claims "Show the claims emitted at each input step").value_parser(value_parser!(bool)))
                 .arg(arg!(-k --show_knowledges "Show the knowledges gathered at each output step").value_parser(value_parser!(bool)))
                 .arg(arg!(-r --show_raw "Show the computed term as raw hex (eg. for use with netcat)").value_parser(value_parser!(bool)))
-                .arg(arg!(-j --json "Export trace execution as JSON").value_parser(value_parser!(bool))),
+                .arg(arg!(-p --differential_post_computations "Evaluate the post execution terms ued in differential fuzzing").value_parser(value_parser!(bool)))
```
nataelbaffou: typo
```suggestion
                .arg(arg!(-p --differential_post_computations "Evaluate the post execution terms used in differential fuzzing").value_parser(value_parser!(bool)))
```

puffin/src/differential.rs
```diff
@@ -0,0 +1,190 @@
+use core::fmt;
+
+use serde::Serialize;
+
+use crate::error::Error;
+use crate::trace::Source;
+
+/// A difference between to exection in differential fuzzing
```
nataelbaffou: typo
```suggestion
/// A difference between to execution in differential fuzzing
```

puffin/src/cli.rs
```diff
@@ -218,19 +237,22 @@ where
...
+                    !experiment_path.as_path().exists(),
+                    "Experiment already exists. Consider creating a new experiment."
+                );
+                (true, experiment_path)
             } else {
                 // Case of non-experiment: plain fuzzing, trace executions, etc.
                 (false, env::current_dir().unwrap())
             }
         };
-    if is_experiment {
```
nataelbaffou, line 226: is_experiment variable is not longer is use, you can remove it and all respective returned values (true, false) from the if-else statement

puffin/src/protocol.rs
```diff
@@ -24,6 +26,74 @@ impl<T: 'static> AsAny for T {
...
+}
+
+impl<T, PT: ProtocolTypes> AsBoxedTerm<PT> for T
+where
+    T: Clone + Debug + EvaluatedTerm<PT> + 'static,
+{
+    fn boxed(&self) -> Box<dyn EvaluatedTerm<PT>> {
+        Box::new(self.clone())
+    }
+}
```
nataelbaffou: It seems unused as the "boxed" method was already is "EvaluatedTerm" trait. Did you want to replace it ?

puffin/src/trace.rs
```diff
@@ -643,19 +798,27 @@ impl<PT: ProtocolTypes> TraceExecution<PT> {
...
                 }
                 Some(step_knowledges)
+                // TODO : Bump rust version to 1.87 to use `extract_if` instead of mem:swap code
+                // Some(
+                //     ctx.knowledge_store
+                //         .raw_knowledge
+                //         .extract_if(.., |k| k.step == Some(StepNumber::new(trace_number, idx)))
+                //         .map(|k| k.data)
+                //         .collect(),
+                // )
```
nataelbaffou, line 804: @LCBH should we add an issue or should it be fixed in the "update_rustc_cargo_versions" branch ?

puffin/src/trace.rs
```diff
@@ -669,13 +832,40 @@ impl<PT: ProtocolTypes> TraceExecution<PT> {
...
+            }
+            Some(knowledges)
+            // TODO : Bump rust version to 1.87 to use `extract_if` instead of mem:swap code
+            // Some(
+            //     ctx.knowledge_store
+            //         .raw_knowledge
+            //         .extract_if(.., |k| k.step == None)
+            //         .map(|k| (k.source, k.data))
+            //         .collect(),
+            // )
```
nataelbaffou, line 853: (COPY) @LCBH should we add an issue or should it be fixed in the "update_rustc_cargo_versions" branch ?

sshpuffin/Cargo.toml
```diff
@@ -23,6 +23,7 @@ puffin-build = { path = "../puffin-build" }
 extractable-macro = { path = "../extractable-macro" }
 log = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
+comparable = { workspace = true }
```
nataelbaffou: You need the "derive" feature and you get it from the puffin/Cargo.toml. But is it a better practive to also include it here ?

tlspuffin/Cargo.toml
```diff
@@ -354,6 +359,7 @@ hex = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 smallvec = { workspace = true }
 once_cell = { workspace = true }
+comparable = { workspace = true }
```
nataelbaffou: (COPY) You need the "derive" feature and you get it from the puffin/Cargo.toml. But is it a better practive to also include it here ?

nataelbaffou (commented):
# Fixes
- sshpuffin/src/ssh/seeds.rs:183 -> missing 3rd parameter in execute method
  - We don't run the sshpuffin tests in the CI ?
- extractable-macro/tests/tests.rs -> errors with trait implementations

# Other notes
- Should some tests for the differential fuzzing be added ? (I might have missed them)
- Lots of scripts in evaluation-ddyf, maybe add an explanation at the beginning of each one to explain a bit more what it does or an MD file to put all those explanations ? (Kinda like the readme that is the artifact of the publication paper)

Maybe I am being too nitpicky, the work is really great and the code clean.

puffin/src/cli.rs
```diff
@@ -93,7 +96,23 @@ where
...
                 .arg(arg!(-p --port [n] "The client port to connect to, or the server port")
-                    .value_parser(value_parser!(u16).range(1..)))
+                    .value_parser(value_parser!(u16).range(1..))),
+            Command::new("differential-execute")
+                .about("Execute a trace on multiple targets, returning the differences between the executions")
+                .arg(arg!(<first_target> "First PUT"))
+                .arg(arg!(<second_target> "Second PUT"))
+                .arg(arg!(<input> "Input trace"))
+                .arg(arg!(-j --json "Export differences as JSON").value_parser(value_parser!(bool))),
+            Command::new("differential")
```
nataelbaffou, line 107: Maybe change the command to "quick-differential" for consistency

puffin/src/cli.rs
nataelbaffou, line 606: Add "differential" or "quick-differential" here for consistency

puffin/src/cli.rs
aeyno: Technically it is not an experiment (objectives are stored in the `objectives` folder) and you can't provide any experiment title like with quick-experiment. This command is closer to running a differential fuzzing campaign without using the fuzzer without any command.

puffin/src/differential.rs
aeyno: Done

puffin/src/cli.rs
aeyno: Done

puffin/src/protocol.rs
aeyno: Done

puffin/src/trace.rs
aeyno: As discussed IRL, we will do another PR to replace the code when the rust version will be updated

tlspuffin/Cargo.toml
aeyno: Done

sshpuffin/Cargo.toml
aeyno: done

puffin/src/trace.rs
aeyno: duplicate

puffin/src/cli.rs
aeyno: duplicate

puffin/src/cli.rs
aeyno: done

puffin/src/claims.rs
aeyno: done

puffin/src/fuzzer/harness.rs
```diff
@@ -35,9 +38,13 @@ pub fn harness<PB: ProtocolBehavior + 'static>(
         }
     }
     // Execute the trace
-    let runner = Runner::new(put_registry.clone(), Spawner::new(put_registry.clone()));
+    let agent = input.descriptors[0].name;
```
LCBH: Protect this in case descriptors is empty.

puffin/src/fuzzer/harness.rs
```diff
@@ -35,9 +38,13 @@ pub fn harness<PB: ProtocolBehavior + 'static>(
...
+    let runner = Runner::new(
+        put_registry.clone(),
+        Spawner::new(put_registry.clone()).with_mapping(&[(agent, put_descriptor.clone())]),
```
LCBH: Why don't we map all agents from the trace? Here, we only do so for the first in the list.

puffin/src/fuzzer/harness.rs
```diff
@@ -57,6 +64,88 @@ pub fn harness<PB: ProtocolBehavior + 'static>(
     ExitKind::Ok
 }
 
+pub fn differential_harness<PB: ProtocolBehavior + 'static>(
+    put_registry: &PutRegistry<PB>,
+    first_put: &PutDescriptor,
+    second_put: &PutDescriptor,
+    input: &Trace<PB::ProtocolTypes>,
+) -> ExitKind {
+    let agent = input.descriptors[0].name;
```
LCBH: Same here

puffin/src/protocol.rs
```diff
@@ -178,6 +237,32 @@ pub trait ProtocolTypes:
...
+    /// Terms to evaluate and compare during differential fuzzing
+    /// eg. decryption recipes
+    /// Terms to evaluate and compare during differential fuzzing
+    /// eg. decryption recipes
+    fn differential_fuzzing_terms_to_eval(
+        agents: &Vec<AgentDescriptor<Self::PUTConfig>>,
+    ) -> Vec<crate::algebra::Term<Self>>;
+
+    /// Replace trace's agent protocol config with parameters
+    /// that unifomize the behavior of the PUTs
```
LCBH: typo

puffin/src/protocol.rs
```diff
@@ -178,6 +237,32 @@ pub trait ProtocolTypes:
...
+    /// Allowed Knowlege's TypeIds when comparing knowledges stores
+    fn differential_fuzzing_whitelist() -> Option<Vec<TypeId>>;
+
+    /// Claims' TypeIds to blacklist when comparing knowledges stores
+    fn differential_fuzzing_claims_blacklist() -> Option<Vec<TypeId>>;
+
+    /// Terms to evaluate and compare during differential fuzzing
+    /// eg. decryption recipes
+    /// Terms to evaluate and compare during differential fuzzing
+    /// eg. decryption recipes
```
LCBH, line 248: typo

puffin/src/protocol.rs
```diff
@@ -293,3 +397,29 @@ macro_rules! atom_extract_knowledge {
...
+        impl $crate::protocol::CompareKnowledge<$protocol_type> for $extract_type {
+            fn find_differences(
+                &self,
+                other: &dyn EvaluatedTerm<$protocol_type>,
+                diffs: &mut Vec<$crate::differential::TraceDifference>,
+                knowledge_num: usize,
+            ) {
+                match other.as_any().downcast_ref::<$extract_type>() {
+                    Some(_) => {
+                        todo!("Comparable for {}", other.type_name());
```
LCBH: Do we really want to panic here? Don't we want log::err instead?

puffin/src/protocol.rs
```diff
@@ -178,6 +237,32 @@ pub trait ProtocolTypes:
...
+    /// eg. decryption recipes
+    fn differential_fuzzing_terms_to_eval(
+        agents: &Vec<AgentDescriptor<Self::PUTConfig>>,
+    ) -> Vec<crate::algebra::Term<Self>>;
+
+    /// Replace trace's agent protocol config with parameters
+    /// that unifomize the behavior of the PUTs
+    fn differential_fuzzing_uniformise_put_config(trace: Trace<Self>) -> Trace<Self>;
+
+    /// Check wether a difference should be kept
```
LCBH: typo

puffin/Cargo.toml
```diff
@@ -25,6 +25,11 @@ with-min-scheduler = []
 
 introspection = ["libafl/introspection"]
 
+ddyf-disable-status = []
```
LCBH, line 29: Add a one-liner comment to explain this is for ablation study

puffin/src/claims.rs
```diff
@@ -94,6 +108,96 @@ impl<C: Claim> ClaimList<C> {
...
+        let mut diffs = vec![];
+
+        // for each agent, compare the claims of both lists and find differences in types and inner
+        // values
+        for k in keys.iter().dedup() {
+            let empty = vec![];
+            let s = self_claims_filtered.get(k).unwrap_or(&empty);
+            let o = other_claims_filtered.get(k).unwrap_or(&empty);
+
+            for i in 0..usize::max(s.len(), o.len()) {
```
LCBH, line 162: Can't we simply apply Comparable on the two lists? (hmm not sure actually since we treat differently the case different content from the case missing content...)

puffin/src/codec.rs
```diff
@@ -417,3 +417,10 @@ impl<T: Codec + VecCodecWoSize> Codec for Vec<T> {
         Some(ret)
     }
 }
+
+pub fn compare_encoding<X: Codec, Y: Codec>(x: &X, y: &Y) -> std::cmp::Ordering {
+    match Codec::get_encoding(x) <= Codec::get_encoding(y) {
+        true => std::cmp::Ordering::Less,
+        false => std::cmp::Ordering::Greater,
```
LCBH: Why no Equal case? Seems to violate the Ord contract

puffin/src/put_registry.rs
```diff
@@ -46,6 +46,10 @@ impl<PB> PutRegistry<PB> {
     pub fn default_put_name(&self) -> &str {
         &self.default_put.factory
     }
+
+    pub fn default_put_descriptor(&self) -> PutDescriptor {
+        PutDescriptor::new(self.default_put_name(), PutOptions::empty())
```
LCBH: Use `self.default_put.options` rather. than `PutOptions::empty()`?!

puffin/src/differential.rs
```diff
@@ -0,0 +1,190 @@
...
+pub enum KnowledgeDiff {
+    /// The two knowledges compared have a different rust type
+    DifferentTypes {
+        index: usize,
+        first_type: String,
+        second_type: String,
+        first_source: Source,
+        second_source: Source,
+    },
+    /// The two knowledges have the same type but there internal values are different
```
LCBH: typo

puffin/src/cli.rs
```diff
@@ -217,20 +236,29 @@ where
...
+                experiment_path
             } else {
                 // Case of non-experiment: plain fuzzing, trace executions, etc.
-                (false, env::current_dir().unwrap())
+                env::current_dir().unwrap()
             }
         };
-    if is_experiment {
-        log::info!("Running in experiment mode");
-        config.is_experiment = is_experiment;
```
LCBH, line 228: We never set config.is_experiment to true now, which is needed for correct logging (see where it was used). Was this intentional?

puffin/src/cli.rs
```diff
@@ -217,20 +236,29 @@ where
                     !experiment_path.as_path().exists(),
                     "Experiment already exists. Consider creating a new experiment."
                 );
-                (true, experiment_path)
+                experiment_path
```
LCBH: Since the two branches have the same code, merge them.

puffin/src/cli.rs
```diff
@@ -434,15 +484,79 @@ where
...
+
+        if let Err((available_puts, non_available_puts)) =
+            check_if_puts_exist(&put_registry, &[first_put, second_put])
+        {
+            log::error!("PUT not found : {}", non_available_puts.join(","));
+            log::error!("Available PUTs: {}", available_puts.join(","));
+            return ExitCode::FAILURE;
+        }
+
+        let agent = trace.descriptors[0].name;
```
LCBH: Better deal with cases where the list is empty.

puffin/src/cli.rs
```diff
@@ -434,15 +484,79 @@ where
...
+
+        let runner = DifferentialRunner::new(
+            put_registry.clone(),
+            Spawner::new(put_registry.clone())
+                .with_mapping(&[(agent, PutDescriptor::new(first_put, PutOptions::empty()))]),
```
LCBH: Why only the first agent of the trace is set?

puffin/src/cli.rs
```diff
@@ -434,15 +484,79 @@ where
...
+            Spawner::new(put_registry)
+                .with_mapping(&[(agent, PutDescriptor::new(second_put, PutOptions::empty()))]),
+        );
+
+        return match runner.execute(trace, &mut 0, false) {
```
LCBH: Is the `false` intended here?

puffin/src/execution.rs
```diff
@@ -8,27 +9,39 @@ use nix::sys::wait::{waitpid, WaitPidFlag};
...
     where
         T: AsRef<Trace<<Self::PB as ProtocolBehavior>::ProtocolTypes>>,
     {
-        Self::execute_config(self, trace, ConfigTrace::default(), executed_until)
+        Self::execute_config(
+            self,
+            trace,
+            ConfigTrace::default(),
+            executed_until,
+            check_security_violation,
```
LCBH: Wouldn't be cleaner to includer this into ConfigTrace instead?

LCBH: Hmm, would need to add this parameter to `execute` as well then. To be discussed.

puffin/src/execution.rs
```diff
@@ -138,6 +164,179 @@ impl<T: TraceRunner + Clone> TraceRunner for &ForkedRunner<T> {
...
+        }
+    }
+}
+
+impl<PB: ProtocolBehavior> TraceRunner for &DifferentialRunner<PB> {
+    type E = Error;
+    type PB = PB;
+    type R = TraceContext<Self::PB>;
+
+    /// Execute `trace` on the two PUT of `self`, return `Ok` if both execution exhibit no
```
LCBH: typos

puffin/src/execution.rs
```diff
@@ -138,6 +164,179 @@ impl<T: TraceRunner + Clone> TraceRunner for &ForkedRunner<T> {
...
+                .as_error());
+            }
+            _ => (),
+        }
+
+        // Maximum executed step on the two PUTs
+        *executed_until = usize::max(first_ctx.executed_until, second_ctx.executed_until);
+
+        //check if we have security claim violation (if security claims are enabled)
+        let mut diff = match (&first_trace_status, &second_trace_status) {
```
LCBH: Is it intended to check this only if there is no status difference? I feel that security claim violations are "more severe" and less prone to false positives....

puffin/src/execution.rs
```diff
@@ -138,6 +164,179 @@ impl<T: TraceRunner + Clone> TraceRunner for &ForkedRunner<T> {
...
+        diff = diff
+            .into_iter()
+            .filter(<<PB as ProtocolBehavior>::ProtocolTypes as ProtocolTypes>::differential_fuzzing_filter_diff)
+            .collect();
+
+        if !diff.is_empty() {
+            return Err(Error::Difference(diff));
+        }
+
+        Ok(first_ctx)
```
LCBH: You never want to access both contexts? I reckon that you already checked that they are both "the same" but this still allows some differences that are filtered out.

puffin/src/execution.rs
```diff
@@ -138,6 +164,179 @@ impl<T: TraceRunner + Clone> TraceRunner for &ForkedRunner<T> {
...
+        executed_until: &mut usize,
+        check_security_violation: bool,
+    ) -> Result<Self::R, Self::E>
+    where
+        T: AsRef<Trace<<Self::PB as ProtocolBehavior>::ProtocolTypes>>,
+    {
+        self.execute_config(
+            trace,
+            ConfigTrace {
+                with_bit_level: false,
```
LCBH: Why not letting the user choose this. It would be nice to do diff fuzzing with havoc mutations enabled.

LCBH: See how it is done for the other implem of `execute`.

puffin/src/trace.rs
```diff
@@ -258,6 +261,92 @@ impl<PT: ProtocolTypes> KnowledgeStore<PT> {
...
+            .count();
+
+        match differences.is_empty() {
+            false => Err(differences),
+            true => Ok(()),
+        }
+    }
+}
+
+/// Should a specific knowledge be filtered out
```
LCBH: "should be kept (not filtered out)" right?

puffin/src/trace.rs
```diff
@@ -258,6 +261,92 @@ impl<PT: ProtocolTypes> KnowledgeStore<PT> {
...
+    knowledge: &Knowledge<PT>,
+    whitelist: &Option<Vec<TypeId>>,
+    blacklist: &Option<Vec<TypeId>>,
+) -> bool {
+    if whitelist.is_none() && blacklist.is_none() {
+        return true;
+    }
+
+    if let Some(list) = whitelist {
+        if !list.iter().any(|x| x == &knowledge.data.type_id()) {
```
LCBH, line 335: Hmmm, I am confused. If it is in the white list --> it is always kept. If not, it is kept only if it is not in the blacklist. No?

puffin/src/trace.rs
LCBH: This will be doable now.

puffin/src/trace.rs
```diff
@@ -643,19 +798,27 @@ impl<PT: ProtocolTypes> TraceExecution<PT> {
...
-                    if c.get_step() == Some(StepNumber::new(trace_number, idx)) {
-                        step_claims.push(c.inner());
-                    }
-                }
-                Some(step_claims)
+                Some(
+                    ctx.claims
+                        .deref_borrow()
+                        .iter()
+                        .filter(|c| c.get_step() == Some(StepNumber::new(trace_number, idx)))
```
LCBH, line 815: This whole block and this filtering needs documentation

LCBH: Actually yhe whole function would need better documentation.

puffin/src/trace.rs
```diff
@@ -669,13 +832,40 @@ impl<PT: ProtocolTypes> TraceExecution<PT> {
...
+            None
+        };
+
         Self {
             agents: trace.descriptors.clone(),
             number_of_steps: trace.steps.len(),
             executed_until: ctx.executed_until,
             steps,
             // right now we exclude prior traces
             prior_traces: Vec::new(),
```
LCBH, line 865: Why not just putting here the prior traces from the trace?

puffin/src/protocol.rs
```diff
@@ -293,3 +397,29 @@ macro_rules! atom_extract_knowledge {
         }
     };
 }
+
+#[macro_export]
+macro_rules! dummy_compare {
```
LCBH: Is this used somewhere (I cannot find any occurence)

puffin/src/protocol.rs
```diff
@@ -293,3 +397,29 @@ macro_rules! atom_extract_knowledge {
...
+    ($protocol_type:ty, $extract_type:ty) => {
+        impl $crate::protocol::CompareKnowledge<$protocol_type> for $extract_type {
+            fn find_differences(
```
LCBH: Wrong function signature!

puffin/src/cli.rs
LCBH: Another occurence of this in this file

puffin/src/trace.rs
```diff
@@ -258,6 +261,92 @@ impl<PT: ProtocolTypes> KnowledgeStore<PT> {
...
+                source: &Source::Label(None),
+                matcher: None,
+                data: &(),
+            }))
+        }
+
+        log::trace!("Comparing knowledge stores");
+        let _ = std::iter::zip(first_store, second_store)
+            .enumerate()
+            .map(|(idx, (x, y))| {
```
LCBH: Rather uses `.for_each()`?

LCBH (changes requested):

tlspuffin/harness/openssl/src/put.c
```diff
@@ -443,10 +443,6 @@ static RESULT get_result(AGENT agent, int retcode, bool allow_would_block)
     case SSL_ERROR_WANT_READ:
     case SSL_ERROR_WANT_WRITE:
         error_type = strdup("IO_WOULD_BLOCK");
-        if (!allow_would_block)
```
LCBH, line 446: This followed quite a lot of discussions and was not trivial at all. It would be nice to document a bit the rationale. Ideally, add a documentation at the start of the case distinctions with the rationale.

puffin/src/cli.rs
```diff
@@ -217,20 +236,29 @@ where
...
             } else {
                 // Case of non-experiment: plain fuzzing, trace executions, etc.
-                (false, env::current_dir().unwrap())
+                env::current_dir().unwrap()
             }
         };
-    if is_experiment {
-        log::info!("Running in experiment mode");
-        config.is_experiment = is_experiment;
-        config.corpus_dir = base_directory.join("corpus");
```
nataelbaffou, line 229: Il faut garder ces configs pour tout avoir dans l'experiment en cours et non à la racine du projet

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -172,9 +173,11 @@ impl From<[u8; 32]> for Random {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Comparable)]
 pub struct SessionID {
+    #[comparable_ignore]
```
LCBH: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -357,12 +361,13 @@ impl ServerNamePayload {
...
 }
 
-#[derive(Clone, Debug, Extractable)]
+#[derive(Clone, Debug, Extractable, Comparable, PartialEq)]
 #[extractable(TLSProtocolTypes)]
 pub struct ServerName {
     #[extractable_ignore]
     pub typ: ServerNameType,
     #[extractable_ignore]
+    #[comparable_ignore]
```
LCBH, line 369: [Compare] Just for the discussion: do we agree that we could theoretically make sure to configure agents so that this field should be aligned? (If yes, just add a comment saying this)

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -554,9 +562,11 @@ impl codec::Codec for PresharedKeyOffer {
 // --- RFC6066 certificate status request ---
 type ResponderIDs = VecU16OfPayloadU16;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Comparable)]
 pub struct OCSPCertificateStatusRequest {
+    #[comparable_ignore]
     pub responder_ids: ResponderIDs,
+    #[comparable_ignore]
```
LCBH: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -630,10 +640,10 @@ declare_u8_vec!(PSKKeyExchangeModes, PSKKeyExchangeMode);
 declare_u16_vec!(KeyShareEntries, KeyShareEntry);
 declare_u8_vec!(ProtocolVersions, ProtocolVersion);
 
-#[derive(Debug, Clone, Extractable)]
+#[derive(Debug, Clone, Extractable, PartialEq, Comparable)]
 #[extractable(TLSProtocolTypes)]
 pub enum ClientExtension {
-    ECPointFormats(ECPointFormatList),
+    ECPointFormats(#[comparable_ignore] ECPointFormatList),
```
LCBH, line 645: [Compare] Just for the discussion: do we agree that we could theoretically make sure to configure agents so that this field should be aligned? (If yes, just add a comment saying this)

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -813,16 +823,16 @@ impl ClientExtension {
...
 pub enum ClientSessionTicket {
     Request,
     Offer(#[extractable_ignore] Payload),
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Comparable)]
 pub enum ServerExtension {
-    ECPointFormats(ECPointFormatList),
+    ECPointFormats(#[comparable_ignore] ECPointFormatList),
```
LCBH, line 835: [Compare] Just for the discussion: do we agree that we could theoretically make sure to configure agents so that this field should be aligned? (If yes, just add a comment saying this)

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -951,14 +961,25 @@ declare_u16_vec_empty!(ClientExtensions, ClientExtension);
 declare_u16_vec!(CipherSuites, CipherSuite);
 declare_u8_vec!(Compressions, Compression);
 
-#[derive(Debug, Clone, Extractable)]
+#[derive(Debug, Clone, Extractable, Comparable)]
 #[extractable(TLSProtocolTypes)]
 pub struct ClientHelloPayload {
+    #[comparable_synthetic {
+        let sorted_cipher_suites = |x: &Self| -> std::collections::HashMap<CipherSuite, ()> { let ciphers : std::collections::HashMap<CipherSuite, ()>= x.cipher_suites.0.clone().into_iter().map(|c| (c,())).collect(); ciphers };
```
LCBH: [compare] why removing duplicates? Don't we want to compare as multisets? (as for extensions)

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -1748,12 +1792,13 @@ impl codec::Codec for ClientECDHParams {
     }
 }
 
-#[derive(Debug, Clone, Extractable)]
+#[derive(Debug, Clone, Extractable, Comparable)]
 #[extractable(TLSProtocolTypes)]
 pub struct ServerECDHParams {
     #[extractable_ignore]
     pub curve_params: ECParameters,
```
LCBH, line 1800: [Compare] Just for the discussion: do we agree that we could theoretically make sure to configure agents so that this field should be aligned? (If yes, just add a comment saying this)

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -2059,10 +2104,12 @@ impl CertificateRequestPayloadTLS13 {
 }
 
 // -- NewSessionTicket --
-#[derive(Debug, Clone, Extractable)]
+#[derive(Debug, Clone, Extractable, Comparable)]
 #[extractable(TLSProtocolTypes)]
 pub struct NewSessionTicketPayload {
+    #[comparable_ignore]
```
LCBH, line 2111: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/tls/rustls/msgs/message.rs
```diff
@@ -53,6 +54,24 @@ impl codec::CodecP for MessagePayload {
         )))
     }
 }
+
+impl PartialEq for MessagePayload {
```
LCBH, line 58: Why do we need a custom PartialEq implem here?

tlspuffin/src/tls/rustls/hash_hs.rs
```diff
@@ -75,9 +76,10 @@ impl HandshakeHashBuffer {
 ///
 /// For client auth, we also need to buffer all the messages.
 /// This is disabled in cases where client auth is not possible.
-#[derive(Clone)]
+#[derive(Clone, Comparable)]
 pub struct HandshakeHash {
     /// None before we know what hash function we're using
+    #[comparable_ignore]
```
LCBH, line 82: [Compare] Wait: why don't we have to ignore the other fields (like `override_buffer`)?

tlspuffin/src/tls/fn_transcript.rs
```diff
@@ -36,3 +38,35 @@ fn _fn_transcript<T: Transcript>(claim: &T) -> Result<HandshakeHash, FnError> {
     let hash = HandshakeHash::new_override(Vec::from(claim.as_slice()), algorithm);
     Ok(hash)
 }
+
+pub fn fn_finished_get_cipher(claim: &Finished) -> Result<CipherSuite, FnError> {
```
LCBH, line 43: Please add somewhere that these function symbols are never meant to be attacker symbols (from the signature). We should ideally avoid using `fn_XX` for such functions.

tlspuffin/src/tls/fn_transcript.rs
```diff
@@ -36,3 +38,35 @@ fn _fn_transcript<T: Transcript>(claim: &T) -> Result<HandshakeHash, FnError> {
...
+    let mut payload = [0; 32];
+    for (idx, v) in claim.server_random.iter().enumerate() {
+        payload[idx] = *v;
+    }
+    Ok(Random(payload))
+}
+
+pub fn fn_finished_get_client_random(claim: &Finished) -> Result<Random, FnError> {
+    let mut payload = [0; 32];
+    for (idx, v) in claim.client_random.iter().enumerate() {
```
LCBH: No better way to copy a SmallVec?

tlspuffin/src/tls/fn_transcript.rs
```diff
@@ -36,3 +38,35 @@ fn _fn_transcript<T: Transcript>(claim: &T) -> Result<HandshakeHash, FnError> {
...
+
+pub fn fn_finished_get_handshake_secret(claim: &Finished) -> Result<Vec<u8>, FnError> {
+    let secret = Vec::from(claim.handshake_secret.as_slice());
+
+    Ok(secret)
+}
+
+pub fn fn_finished_get_server_random(claim: &Finished) -> Result<Random, FnError> {
+    let mut payload = [0; 32];
+    for (idx, v) in claim.server_random.iter().enumerate() {
```
LCBH: No better way to copy a SmallVec?

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -158,6 +166,39 @@ pub fn fn_decrypt_handshake_flight(
     Ok(decrypted_flight)
 }
 
+pub fn fn_decrypt_handshake_flight_with_secret(
```
LCBH, line 175: Add a Rust doc

LCBH (changes requested):

tlspuffin/src/claims.rs
```diff
@@ -60,8 +61,9 @@ impl codec::Codec for TranscriptClientHello {
 
 dummy_extract_knowledge!(TLSProtocolTypes, TranscriptClientHello);
 
-#[derive(Debug, Clone)]
-pub struct TranscriptPartialClientHello(pub TlsTranscript);
+#[derive(Debug, Clone, Comparable, PartialEq)]
+pub struct TranscriptPartialClientHello(#[comparable_ignore] pub TlsTranscript);
```
LCBH, line 65: [Compare] Why ignoring this? Shouldn't we compare the length at least? (in TlsTranscript) --> same for other structs in this file

tlspuffin/src/claims.rs
```diff
@@ -169,41 +174,56 @@ pub trait Transcript {
...
 pub struct Certificate;
 dummy_extract_knowledge_codec!(TLSProtocolTypes, Certificate);
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Comparable, PartialEq)]
 pub struct CertificateVerify;
 dummy_extract_knowledge_codec!(TLSProtocolTypes, CertificateVerify);
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Comparable, PartialEq)]
 pub struct Finished {
+    #[comparable_ignore]
```
LCBH, line 192: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/claims.rs
```diff
@@ -169,41 +174,56 @@ pub trait Transcript {
...
+    // }; }]
+    #[comparable_ignore]
     pub early_secret: SmallVec<[u8; 32]>,
+    #[comparable_ignore]
     pub handshake_secret: SmallVec<[u8; 32]>,
+    #[comparable_ignore]
     pub master_secret: SmallVec<[u8; 32]>,
 
     pub chosen_cipher: u16,
+    #[comparable_ignore]
```
LCBH, line 224: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/claims.rs
```diff
@@ -169,41 +174,56 @@ pub trait Transcript {
...
     pub available_ciphers: SmallVec<[u16; 20]>,
 
+    #[comparable_ignore]
```
LCBH, line 232: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/claims.rs
```diff
@@ -169,41 +174,56 @@ pub trait Transcript {
...
     pub signature_algorithm: i32,
+    #[comparable_ignore]
```
LCBH, line 234: [Compare] Why ignoring this? Is this randomized? Different for different PUTs?

tlspuffin/src/claims.rs
```diff
@@ -251,18 +271,19 @@ pub enum ClaimDataMessage {
...
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Comparable, PartialEq)]
 pub struct TlsClaim {
     pub agent_name: AgentName,
     pub origin: AgentType,
     pub protocol_version: TLSVersion,
     pub data: ClaimData,
+    #[comparable_ignore]
```
LCBH, line 294: [Compare] Why ignoring this? Is it because we already test the claim step anyway? But then we could also compare this field?!

LCBH (changes requested):

tlspuffin/src/query.rs
```diff
@@ -12,6 +12,10 @@ pub enum TlsQueryMatcher {
     Handshake(Option<HandshakeType>),
     ApplicationData,
     Heartbeat,
+    ClientHelloFlight,
```
LCBH, line 16: Add some documentation: For example, this is supposed to match any flight starting with a client hello

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
     fn signature() -> &'static Signature<Self> {
         &TLS_SIGNATURE
     }
+
+    fn differential_fuzzing_blacklist() -> Option<Vec<TypeId>> {
```
LCBH: Question: None: is this intended? Why did we need a black list then?

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
...
+
+    fn differential_fuzzing_terms_to_eval(
+        agents: &Vec<AgentDescriptor<Self::PUTConfig>>,
+    ) -> Vec<puffin::algebra::Term<Self>> {
+        let mut is_server = false;
+        let mut server = AgentName::new();
+        let mut is_client = false;
+        let mut client = AgentName::new();
+
+        for agent in agents {
```
LCBH, line 423: Would be nice to document this

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
...
+                (fn_finished_get_cipher(((client, 0)))),
+                (fn_finished_get_handshake_secret(((client, 0))))
+            )
+            });
+        }
+
+        terms
+    }
+
+    fn differential_fuzzing_claims_blacklist() -> Option<Vec<TypeId>> {
```
LCBH, line 466: [Compare] Why all of these transcripts? Always different?

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
...
+            TypeId::of::<TranscriptCertificate>(),
+            TypeId::of::<TranscriptClientFinished>(),
+            TypeId::of::<TranscriptClientHello>(),
+            TypeId::of::<TranscriptPartialClientHello>(),
+            TypeId::of::<TranscriptServerFinished>(),
+            TypeId::of::<TranscriptServerHello>(),
+        ])
+    }
+
+    fn differential_fuzzing_uniformise_put_config(mut trace: Trace<Self>) -> Trace<Self> {
```
LCBH: Why taking a `mut` on the input? (the signature is weird)

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
...
+        for agent in trace.descriptors.iter_mut() {
```
LCBH: Document somewhere (ideally on the official doc) for which vendors this works.

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
...
+                "TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256:TLS_CHACHA20_POLY1305_SHA256",
+            );
+            agent.protocol_config.groups = Some(String::from("P-256:P-384"));
+        }
+
+        for t in trace.prior_traces.iter_mut() {
+            *t = Self::differential_fuzzing_uniformise_put_config(t.to_owned());
+        }
+
+        trace
```
LCBH: Is there a way to test if this has worked (when executing the trace? Is set-cipher capable of detecting failures?)

tlspuffin/src/protocol.rs
LCBH: Also document here why we chose these ciphers.

tlspuffin/src/protocol.rs
```diff
@@ -360,6 +405,110 @@ impl ProtocolTypes for TLSProtocolTypes {
...
+    fn differential_fuzzing_filter_diff(diff: &TraceDifference) -> bool {
+        if let TraceDifference::Knowledges(puffin::differential::KnowledgeDiff::InnerDifference {
+            index: _,
+            type_name: _,
+            diff,
+            source: _,
+        }) = diff
+        {
+            // OpenSSL sends an optional EllipticCurves extension when renegociating
+            if diff.contains("EncryptedExtensionsChange([Removed(0, Unknown(UnknownExtensionDesc { typ: EllipticCurves }))])") {
```
LCBH: This does not resist changes in struct removing or structure modifications. Is there a way to make this test more robust?

tlspuffin/Cargo.toml
```diff
@@ -340,6 +340,11 @@ enable-guards = []
 # utility functions for testing tlspuffin
 test-utils = ["tempfile"]
 
+ddyf-disable-status = ["puffin/ddyf-disable-status"]
```
LCBH, line 344: please document

tools/differential/list_buckets.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
LCBH, line 1: This script needs documentation

tools/differential/empty_buckets.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
LCBH, line 1: This script needs documentation

LCBH (changes requested):

puffin/src/fuzzer/harness.rs
aeyno: done

puffin/src/fuzzer/harness.rs
aeyno: done

puffin/src/fuzzer/harness.rs
aeyno: done

puffin/src/protocol.rs
aeyno: done

puffin/src/protocol.rs
aeyno: Done

puffin/Cargo.toml
aeyno: Done

puffin/src/codec.rs
aeyno: Changed to support equal case

puffin/src/differential.rs
aeyno: done

puffin/src/codec.rs
```diff
@@ -417,3 +417,10 @@ impl<T: Codec + VecCodecWoSize> Codec for Vec<T> {
...
+    }
```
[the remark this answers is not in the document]
aeyno: done

tlspuffin/Cargo.toml
aeyno: done

tlspuffin/src/protocol.rs
aeyno: done

puffin/src/put_registry.rs
aeyno: done

puffin/src/protocol.rs
aeyno: This macro was not used anymore so I removed it

puffin/src/cli.rs
aeyno: done

puffin/src/cli.rs
aeyno: done

puffin/src/cli.rs
aeyno: done

puffin/src/execution.rs
aeyno: done

puffin/src/execution.rs
aeyno: We just abort the comparison as soon as we detect a status difference for performance reasons , but this kind of differences should be catched during the trace analysis.

puffin/src/protocol.rs
aeyno: Indeed it was not used anymore so I removed it

puffin/src/claims.rs
aeyno: We could but this increases the readability of the differences: else we would only have a big data structure with all the differences (inner diff and different types) for all the claims.

puffin/src/cli.rs
aeyno: This looks like an error during a rebase. I put it back.

puffin/src/cli.rs
aeyno: I added `--enable-security-oracle` flag to enable the security oracle which is disabled by default in differential mode.

tlspuffin/src/protocol.rs
aeyno: Yes all transcripts are different since they all depend on CH and SH which contains random values.

tlspuffin/src/claims.rs
aeyno: It was different for OpenSSL and wolfSSL, but I think you introduced the necessary API to changes sigalgs in [!468](https://github.com/tlspuffin/tlspuffin/pull/468). So this could probably be removed when merged.

tlspuffin/src/claims.rs
aeyno: list of ciphers are different for TLS 1.2, openssl shows 5 (3 TLS 1.3 and two TLS 1.2) while wolfssl shows only the two TLS 1.2 ciphers. So this is an API difference.

tlspuffin/src/claims.rs
aeyno: This is ignored because it introduces duplicates Finished claims on some PUTs: ignoring this fields makes the oracle deduplicate all adjacent Finished while not ignoring it causes some PUTs to have this claims multiple times with and without `outbound=true`

puffin/src/execution.rs
aeyno: Changed to return both

puffin/src/execution.rs
aeyno: Changed.

puffin/src/trace.rs
aeyno: Indeed. Changed.

puffin/src/trace.rs
aeyno: Done

puffin/src/cli.rs
aeyno: done

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Since we already ignore SessionIDs in CH and SH, I removed the ignore here

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Yes probably, I added the comment.

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: I tried to remove the ignores and I don't have differences on the seed so I am not sure why I ignored them

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Done

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Done

puffin/src/trace.rs
aeyno: I don't think that the blacklist is really useful, I will probably remove it

puffin/src/protocol.rs
aeyno: This macro has been removed

puffin/src/trace.rs
aeyno: Done

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Ideed, I changed that

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Done

tlspuffin/src/tls/rustls/msgs/message.rs
aeyno: I think that is to avoid having to derive `PartialEq` on all the underlying types (basically all rustls types)

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Yes, the lifetime and ticket payloads are PUT dependant. I don't think it is possible to change this value for OpenSSL

tlspuffin/src/tls/rustls/hash_hs.rs
aeyno: The reason why this field is ignored is that now `digest::Context` doesn't have to implement Comparable.

tlspuffin/src/tls/fn_transcript.rs
aeyno: Done

tlspuffin/src/tls/fn_transcript.rs
aeyno: Done.

tlspuffin/src/tls/fn_transcript.rs
aeyno: Done.

tlspuffin/src/tls/fn_utils.rs
aeyno: Done.

puffin/src/trace.rs
aeyno: Discussed offline. This requires some changes to the way the knowledges are stored to correctly map them to the correct prior trace. I opened an issue: https://github.com/tlspuffin/tlspuffin/issues/476

tlspuffin/src/claims.rs
aeyno: We ignore the step number because when a claim is present multiple times at multiple steps we can't deduplicate it if they have different step numbers. So we ignore the step number to focus on the succession of claims rather than when they were emitted.

tlspuffin/src/claims.rs
aeyno: It was different for OpenSSL and wolfSSL, but I think you introduced the necessary API to changes sigalgs in https://github.com/tlspuffin/tlspuffin/pull/468. So this could probably be removed when merged.

tlspuffin/src/query.rs
aeyno: Done

tlspuffin/src/protocol.rs
aeyno: Blacklist has been removed

tlspuffin/src/protocol.rs
aeyno: Done.

tlspuffin/src/protocol.rs
aeyno: Done.

tlspuffin/src/protocol.rs
aeyno: set_cipher and set_groups are tested in `tlspuffin/tests/harness.rs`

tlspuffin/src/protocol.rs
aeyno: I changed that to just searching for a difference containing `"UnknownExtensionDesc { typ: EllipticCurves }"`

puffin/src/trace.rs
```diff
@@ -24,13 +24,16 @@ use std::marker::PhantomData;
 use std::mem;
 use std::vec::IntoIter;
 
+use clap::error::Result;
+use comparable::Comparable;
```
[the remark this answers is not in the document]
aeyno: Done.

tools/differential/list_buckets.sh
```diff
@@ -0,0 +1,22 @@
...
+
+total=$(find $folder -maxdepth 1 -type f -regextype posix-egrep -regex '.*\.trace(-[0-9]+)?' | wc -l)
+
+echo "non triaged : $total"
+
+for d in $folder/*; do
+  if [ -d "$d" ];
+    then
+    num=$(find $d -maxdepth 1 -type f -regextype posix-egrep -regex '.*\.trace(-[0-9]+)?' | wc -l)
+    total=$((total+num))
```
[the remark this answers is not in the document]
aeyno: Done.

docs/docs/guides/getting-started/differential-fuzzing.md
```diff
@@ -0,0 +1,169 @@
...
+./target/release/tlspuffin --put openssl340 display-execute -tckp path/to/trace
+```
+
+where `-t` shows the terms of the trace, `-c` show the claims emitted by the PUT, `-k` show the knowledges extracted from the messages and `-p` show the decrypted knowledges.
+
+## Trace triaging
+
+To triage all the traces in the objectives, we provide a Python library to automate grouping traces.
+
+Start by importing the necessary elements from `./tool/differential/diff-analyzer.py`
```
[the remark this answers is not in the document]
aeyno: Done

docs/docs/guides/getting-started/differential-fuzzing.md
```diff
@@ -0,0 +1,169 @@
...
+---
+
+## Differential fuzzing
+
+*Puffin* based fuzzers can do _differential fuzzing_ between two PUTs: each trace is executed on the two PUTs and if their executions are different (execution status, knowledge, claims) the trace is flagged as an objective.
+
+
+## Requirements
+
+In order to do differential fuzzing, the protocol fuzzer should be configured to minimize the false positives, i.e. all randomized fields (nonces, random, secrets, hash, ciphertext, etc.) should be annotated to be ignored during the comparison step. If the protocol offer some cusomization (e.g. cipher selection), the protocol fuzzer must provide a set of common parameters for all the PUTs to minimize the difference. The fuzzer might also provide a set of terms to decipher the encrypted messages send by the PUTs by leveraging the secrets extracted via the claims.
```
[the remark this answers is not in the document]
aeyno: Done.

extractable-macro/tests/tests.rs
```diff
@@ -35,9 +36,37 @@ impl ProtocolTypes for TestProtocolTypes {
     fn signature() -> &'static Signature<Self> {
         &TEST_SIGNATURE
     }
+
+    fn differential_fuzzing_blacklist() -> Option<Vec<TypeId>> {
+        todo!()
+    }
```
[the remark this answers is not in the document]
aeyno: Blacklists have been removed.

tlspuffin/src/claims.rs
aeyno: All transcripts are ignored since their content must be different.

tools/differential/empty_buckets.sh
aeyno: Done.

tools/differential/list_buckets.sh
aeyno: Done.

puffin/src/execution.rs
aeyno: TODO: make a security violation obvious to be sure to correctly triage it

puffin/src/execution.rs
aeyno: Done.

puffin/src/execution.rs
aeyno: Done.

tlspuffin/harness/openssl/src/put.c
aeyno: Done

aeyno:
@LCBH can you allow merging this PR ?

LCBH (approved):

PR #346, opened 2024-11-05
Removing `VariableData` to use `EvaluatedTerm` instead

aeyno:
This PR removes the `VariableData` trait in favor of `EvaluatedTerm` to simplify the code and remove the conversions between those two types.


michaelmera (approved):
LGTM

michaelmera:
@aeyno Current `main` branch has `Extractable` vs `EvalutedTerm` in this PR. Which should it be?

PR #352, opened 2024-12-12
Adding  flag to select a PUT from the PUT registry

aeyno:
This PR adds a `--put` CLI flag to select the PUT from the registry to use for fuzzing or executing a trace.

### DIFF
--- a/puffin/src/cli.rs
+++ b/puffin/src/cli.rs
@@ -116,6 +116,19 @@ where
     let without_dy_mutations = matches.get_flag("wo-dy");
     let target_put: Option<&String> = matches.get_one("put");
 
+    let mut put_registry = put_registry.clone();
+
+    if let Some(name) = target_put {
+        if let Err((available_puts, non_available_puts)) =
+            check_if_puts_exist(&put_registry, &[name])
+        {
+            log::error!("PUT not found : {}", non_available_puts.join(","));
+            log::error!("Available PUTs: {}", available_puts.join(","));
+            return ExitCode::FAILURE;
+        };
+        let _ = put_registry.set_default(name);
+    };
+
     log::info!("Version: {}", puffin::full_version());
     log::info!("Put Versions:");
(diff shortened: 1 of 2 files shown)


puffin/src/cli.rs
```diff
@@ -553,3 +563,29 @@ fn binary_attack<PB: ProtocolBehavior>(
...
+            }
+        })
+        .collect();
+
+    if non_available_puts.is_empty() {
+        true
+    } else {
+        println!("PUT not found : {}", non_available_puts.join(","));
+        println!("Available PUTs: {}", available_puts.join(","));
+        false
```
michaelmera, line 589: Use logging functions instead of `println`.

Also, it's difficult to set a right log level here because the intent of the caller might vary. This suggests that here is not the correct place for these log functions.

How about doing a function `missing_puts` instead, returning the vec of missing PUTs? Then the intent is more clear and in the CLI you already perform a check anyway, which let you group your logging functions together with the `return ExitCode::FAILURE`. Then it's clear that the log level should be `error` in this case.

puffin/src/cli.rs
```diff
@@ -31,6 +31,7 @@ where
         .version(puffin::version())
         .author(crate_authors!())
         .about(title.as_ref().to_owned())
+        .arg(arg!(-T --target <T> "The PUT to use"))
```
michaelmera, line 34: Using `--put` might be more clear?  Anyway at some point we will need to cleanup the CLI so I'm okay with `--target` for the moment. An opinion @LCBH?

michaelmera (changes requested):

puffin/src/cli.rs
LCBH: In favor for `--put` as well.

LCBH (approved):
LGTM

aeyno:
@michaelmera

michaelmera (approved):
LGTM

PR #357, opened 2025-01-28
Move PUT configuration to protocol specific crates

aeyno:
This PR moves protocol specific code from `AgentDescriptor` to protocol crates. This removes the TLS specific code from puffin.
It also adds the ability to change the list of cipher suites used by the PUTs.

### DIFF
--- a/puffin/src/agent.rs
+++ b/puffin/src/agent.rs
@@ -57,7 +57,7 @@ impl From<AgentName> for u8 {
     }
 }
 
-pub trait ProtocolPUTDescriptorConfig:
+pub trait ProtocolDescriptorConfig:
     Default + Debug + Clone + Serialize + Hash + for<'a> Deserialize<'a>
 {
     fn is_reusable_with(&self, other: &Self) -> bool;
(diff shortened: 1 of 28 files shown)


puffin/src/agent.rs
```diff
@@ -61,6 +57,12 @@ impl From<AgentName> for u8 {
     }
 }
 
+pub trait ProtocolPUTDescriptorConfig:
```
LCBH, line 60: add Rust Doc

LCBH: rename to `ProtocolDescriptorConfig`?

tlspuffin/src/protocol.rs
```diff
@@ -809,11 +810,91 @@ impl Matcher for msgs::enums::HandshakeType {
...
+            ..Self::default()
+        };
+
+        AgentDescriptor { name, put_config }
+    }
+}
+
+impl ProtocolPUTDescriptorConfig for TLSPUTDescriptorConfig {
+    fn is_reusable_with(&self, other: &Self) -> bool {
+        self.typ == other.typ && self.tls_version == other.tls_version
```
LCBH, line 875: Don't ypu want to make sure `ciphers` are the same as well?

puffin/src/agent.rs
```diff
@@ -69,96 +71,39 @@ impl From<AgentName> for u8 {
...
-        Self {
-            name: AgentName::first(),
-            tls_version: TLSVersion::V1_3,
-            typ: AgentType::Server,
-            try_reuse: false,
-            client_authentication: false,
-            server_authentication: true,
-        }
-    }
+    pub put_config: C,
```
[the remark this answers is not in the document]
aeyno: done

puffin/src/agent.rs
aeyno: done

tlspuffin/src/protocol.rs
aeyno: Yes, I didn't though about it but indeed if, in the future, we have some cases where we want to have different cipher config for different agents we should check that

aeyno: Done

LCBH (approved):

PR #369, opened 2025-02-14
tlspuffin: allow setting group/curve in PUT

aeyno:


tlspuffin/src/protocol.rs
```diff
@@ -893,6 +895,7 @@ impl Default for TLSDescriptorConfig {
             try_reuse: false,
             typ: AgentType::Server,
             cipher_string: String::from("ALL:!EXPORT:!LOW:!aNULL:!eNULL:!SSLv2"),
+            groups: None,
```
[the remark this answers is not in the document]
LCBH: Got it!

aeyno:
While implementing the unit test to check that we can change the curves, I found out that it was not working for OpenSSL, so I used another API to change the curves consistently. The only downside is that now WolfSSL doesn't support `X25519`, when using `wolfSSL_CTX_set1_groups_list` but it was able to use it when using `wolfSSL_CTX_set_groups`. So I will need to further investigate

aeyno:
Now the test is passing on all PUTs except for wolfssl340 where the test is currently disabled. There is a tracking issue for that : https://github.com/tlspuffin/tlspuffin/issues/375

tlspuffin/src/protocol.rs
```diff
@@ -603,6 +602,79 @@ impl Extractable<TLSProtocolTypes> for ClientHelloPayload {
     }
 }
 
+impl Extractable<TLSProtocolTypes> for ClientExtension {
```
LCBH, line 605: Are those modifications really related to the PR?

tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -40,7 +40,25 @@ macro_rules! declare_u16_vec (
     #[derive(Debug, Clone)]
     pub struct $name(pub Vec<$itemtype>);
 
-    atom_extract_knowledge!(TLSProtocolTypes, $name);
+    impl Extractable<TLSProtocolTypes> for $name {
```
LCBH, line 43: Is this related to the PR?

tlspuffin/src/protocol.rs
```diff
@@ -846,6 +925,8 @@ pub struct TLSDescriptorConfig {
     pub try_reuse: bool,
     /// List of available TLS ciphers
     pub cipher_string: String,
+    /// List of available TLS groups/curves
+    pub groups: Option<String>,
```
LCBH, line 929: By default this is None, so please modify the comment to explain what this field will do then (overwrite the PUT default?)

LCBH (changes requested):

tlspuffin/src/protocol.rs
aeyno: Yes, in order to query the NamedGroup in the unit test, I added the abilty to extract NamedGroup from the client hello.

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: Yes, in order to query the NamedGroup in the unit test, I added the abilty to extract NamedGroup from the client hello.

1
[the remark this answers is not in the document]
aeyno: Done

tlspuffin/src/protocol.rs
aeyno: Done

LCBH (approved):

PR #372, opened 2025-03-03
Improving OpenSSL error handling

aeyno:
This PR improves the error logging of OpenSSl and makes more error non-blocking

### DIFF
--- a/puffin/src/harness.rs
+++ b/puffin/src/harness.rs
@@ -70,6 +70,7 @@ unsafe extern "C" fn make_result(code: RESULT_CODE, description: *const c_char) 
         }),
     });
 
+    log::trace!("PUT error log: {:?}", &result);
     Box::into_raw(result) as *mut _
 }
 
--- a/tlspuffin/harness/openssl/src/put.c
+++ b/tlspuffin/harness/openssl/src/put.c
@@ -375,18 +375,38 @@ static RESULT get_result(AGENT agent, int retcode, bool allow_would_block)
         break;
     case SSL_ERROR_WANT_CONNECT:
         error_type = strdup("SSL_ERROR_WANT_CONNECT");
+        if (!allow_would_block)
+        {
+            res = RESULT_IO_WOULD_BLOCK;
+        }
         break;
     case SSL_ERROR_WANT_ACCEPT:
         error_type = strdup("SSL_ERROR_WANT_ACCEPT");
+        if (!allow_would_block)
+        {
+            res = RESULT_IO_WOULD_BLOCK;
+        }
         break;
     case SSL_ERROR_WANT_X509_LOOKUP:
         error_type = strdup("SSL_ERROR_WANT_X509_LOOKUP");
+        if (!allow_would_block)
+        {
+            res = RESULT_IO_WOULD_BLOCK;
+        }
         break;
     case SSL_ERROR_WANT_ASYNC:
         error_type = strdup("SSL_ERROR_WANT_ASYNC");
+        if (!allow_would_block)
+        {
+            res = RESULT_IO_WOULD_BLOCK;
+        }
         break;
     case SSL_ERROR_WANT_ASYNC_JOB:
         error_type = strdup("SSL_ERROR_WANT_ASYNC_JOB");
+        if (!allow_would_block)
+        {
+            res = RESULT_IO_WOULD_BLOCK;
+        }
         break;
     case SSL_ERROR_WANT_CLIENT_HELLO_CB:
         error_type = strdup("SSL_ERROR_WANT_CLIENT_HELLO_CB");
--- a/tlspuffin/harness/openssl/src/put.c
+++ b/tlspuffin/harness/openssl/src/put.c
@@ -415,13 +415,18 @@ static RESULT get_result(AGENT agent, int retcode, bool allow_would_block)
     char *msg;
     if (strlen(reason) > 0)
     {
-        msg = malloc((strlen(error_type) + strlen(reason) + 3) * sizeof(char));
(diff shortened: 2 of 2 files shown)


tlspuffin/harness/openssl/src/put.c
```diff
@@ -383,18 +357,75 @@ static AGENT make_agent(SSL_CTX *ssl_ctx, const TLS_AGENT_DESCRIPTOR *descriptor
...
     {
     case SSL_ERROR_NONE:
-        return RESULT_OK;
-
+        error_type = strdup("no error");
+        break;
+    case SSL_ERROR_ZERO_RETURN:
+        error_type = strdup("SSL_ERROR_ZERO_RETURN");
+        break;
+    case SSL_ERROR_WANT_CONNECT:
```
LCBH, line 376: looks like this is rather a `IO_WOULD_BLOCK` according to https://docs.openssl.org/1.1.1/man3/SSL_get_error/#return-values
Might also be the case for other types of errors? (I did not check all of them)

tlspuffin/harness/openssl/src/put.c
```diff
@@ -383,18 +357,75 @@ static AGENT make_agent(SSL_CTX *ssl_ctx, const TLS_AGENT_DESCRIPTOR *descriptor
...
+        error_type = strdup("SSL_ERROR_WANT_ASYNC_JOB");
+        break;
+    case SSL_ERROR_WANT_CLIENT_HELLO_CB:
+        error_type = strdup("SSL_ERROR_WANT_CLIENT_HELLO_CB");
+        break;
+    case SSL_ERROR_SYSCALL:
+        error_type = strdup("SSL_ERROR_SYSCALL");
+        break;
+    case SSL_ERROR_SSL:
+        error_type = strdup("SSL_ERROR_SSL");
```
LCBH, line 398: This should be considered equivalent to a crash!
See from https://docs.openssl.org/1.1.1/man3/SSL_get_error/#return-values:
SSL_ERROR_SSL

A non-recoverable, fatal error in the SSL library occurred, usually a protocol error. The OpenSSL error queue contains more information on the error. If this error occurs then no further I/O operations should be performed on the connection and SSL_shutdown() must not be called.

tlspuffin/harness/openssl/src/put.c
```diff
@@ -383,18 +357,75 @@ static AGENT make_agent(SSL_CTX *ssl_ctx, const TLS_AGENT_DESCRIPTOR *descriptor
...
-        return RESULT_ERROR_OTHER;
+        error_type = malloc(32 * sizeof(char));
+        snprintf(error_type, sizeof(error_type), "UNKNOWN SSL ERROR %d", ssl_ecode);
+        res = RESULT_ERROR_OTHER;
+    }
+
+    char *msg;
+    if (strlen(reason) > 0)
+    {
+        msg = malloc((strlen(error_type) + strlen(reason) + 3) * sizeof(char));
```
LCBH, line 418: I suggest you also add the original SSL error `ssl_ecode` in there.

tlspuffin/harness/openssl/src/put.c
```diff
@@ -383,18 +357,75 @@ static AGENT make_agent(SSL_CTX *ssl_ctx, const TLS_AGENT_DESCRIPTOR *descriptor
...
-static RESULT_CODE result_code(AGENT agent, int retcode)
+static RESULT get_result(AGENT agent, int retcode, bool allow_would_block)
 {
     int ssl_ecode = SSL_get_error(agent->ssl, retcode);
+
+    char *reason = get_error_reason();
+    char *error_type;
+    RESULT_CODE res = RESULT_OK;
+
     switch (ssl_ecode)
```
LCBH, line 368: Please link an Issue to this for later investigating further which OpenSSL errors should be considered as objectives. The issue should also document the potential re-design of the types of errors we want to consider in `puffin` (protocol-independent), something like:
 - "Graceful" (all the "warnings) --> just log them,
 - "IO" (IO-related errors or warnings) can stop the trace execution but not an objective --> maybe another execution is needed?
 -  "NonBlocking" (ungraceful error that should be considered as an objective and yet we could continue the trace execution --> it depends, definitely want to continue the execution in differential-mode,
 - "Objective" (ungraceful, considered as objectives) --> stop and flag as objective.

tlspuffin/harness/openssl/src/put.c
aeyno: As discussed, an SSL error is not a crash, since it has been catched by OpenSSL : the server should abort the current connection but is not in a state of DoS. Furthermore, we want this to be non-blocking to see if we can still exchange messages with the PUT once it is in an error state to potentially find bugs.

puffin/src/harness.rs
```diff
@@ -70,6 +70,7 @@ unsafe extern "C" fn make_result(code: RESULT_CODE, description: *const c_char)
         }),
     });
 
+    log::trace!("PUT error log: {:?}", &result);
```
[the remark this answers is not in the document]
aeyno: I'm not sure to understand what you mean

tlspuffin/harness/openssl/src/put.c
aeyno: Done

tlspuffin/harness/openssl/src/put.c
aeyno: As I said previously, I don't think that any error reported by a PUT should be considered as an objective, since it means that the PUT gracefully caught the error. And in this case we probably want to continue fuzzing to see if we can reach a crash or problematic behavior after the PUT returned an error

LCBH (approved):

PR #376, opened 2025-03-09
Add proc macro to automate implementation of Extractable

aeyno:
This PR adds the `Extractable` proc macro to derive `Extractable` trait for protocol types

### DIFF
--- a/extractable-macro/Cargo.toml
+++ b/extractable-macro/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.1"
 authors = ["Tom Gouville"]
 edition = "2021"
 include = ["src/*.rs"]
+description = "This crate provide the Extractable derive macro, that implements the Extractable trait for structs, enum and unions"
 
 [lib]
 proc-macro = true
--- a/extractable-macro/src/lib.rs
+++ b/extractable-macro/src/lib.rs
@@ -3,6 +3,112 @@ use quote::{format_ident, quote};
 use syn::punctuated::Punctuated;
 use syn::{Token, Type};
 
+/// Implements Extractable trait for structs, enums and unions
+///
+/// # Types requirements
+///
+/// All types must implement Clone, Debug and Codec
+///
+/// # Struct
+///
+/// Extractable can be used on struct with named or unnamed fields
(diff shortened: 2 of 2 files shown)


sshpuffin/src/protocol.rs
```diff
@@ -67,30 +69,14 @@ impl From<SshMessage> for SshMessageFlight {
...
-}
-
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Extractable)]
+#[extractable(SshProtocolTypes)]
 pub struct RawSshMessageFlight {
     pub messages: Vec<RawSshMessage>,
 }
 
+impl VecCodecWoSize for RawSshMessage {}
```
LCBH, line 78: This is to get `CodecP`, why do we need it here?

tlspuffin/src/tls/rustls/msgs/heartbeat.rs
```diff
@@ -1,10 +1,14 @@
...
 use crate::tls::rustls::msgs::enums::HeartbeatMessageType;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Extractable)]
+#[extractable(TLSProtocolTypes)]
 pub struct HeartbeatPayload {
+    #[extractable_ignore]
     pub typ: HeartbeatMessageType,
     pub payload: PayloadU16,
     pub fake_length: Option<u16>,
```
LCBH, line 14: I think this was ignored before.
It would be nice to limit this PR to replace previous extractations with the macro but leave more extractions to a future PR. So that evaluation will be easier in the sense that this PR should not impact the fuzzer!

LCBH: Please check this globally.

LCBH (changes requested):

extractable-macro/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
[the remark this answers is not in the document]
aeyno: Done

LCBH (approved):

PR #382, opened 2025-03-14
First draft of OPCUA puffin

vdh-anssi:
This is the new structure of the project OPC UA puffin.
A first attempt to add an extract of rustopcua/types.


Cargo.toml
```diff
@@ -34,48 +36,52 @@ authors = [
...
-tempfile = "3.3.0"
-chrono = "0.4.19"
-criterion = "0.5"
-itertools = "0.13"
-hex = "0.4.3"
-cfg-if = "1.0.0"
-once_cell = "1.12.0"
-dyn-clone = "1.0.17"
-smallvec = "1.8.1"
+anyhow = "1.0"
```
LCBH, line 50: Clean this diff.

opcuapuffin/src/tests.rs
```diff
@@ -0,0 +1,56 @@
...
+use puffin::codec::CodecP;
+
+#[test]
+pub fn client() {
+
+    let max_size = 5000;
+    let mut send_buffer: Vec<u8> = Vec::with_capacity(max_size);
+
+    let hello_message: HelloMessage = fn_hello(
+        &UAString::from("opc.tcp://PenDuick:53530/OPCUA/SimulationServer"),
```
LCBH, line 13: Two points:
1. Construct a `term` rather than a `message`. Evaluate and encode and compare as you do here.
2. To do so, make sure all arguments can be produced with a term. For example here, I don't think there is a term producing a `UAString`.

opcuapuffin/src/tests.rs
```diff
@@ -0,0 +1,56 @@
...
+    let acknowledge_message: AcknowledgeMessage = fn_acknowledge(
+        &max_size,
+        &max_size).unwrap();
+    acknowledge_message.encode(&mut send_buffer);
+    let ack_msg : Vec<u8> = vec![
+    0x41, 0x43, 0x4b, 0x46, 0x1c, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x88, 0x13, 0x00, 0x00,
+    0x88, 0x13, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00];
+    assert_eq!(&send_buffer,  &ack_msg);
+      //println!("acknowledge: {:x?}",  send_buffer);
+}
```
LCBH, line 56: Add a test with open secure channel too.

LCBH (commented):
Some coments, to discuss.

PR #383, opened 2025-03-18
wolfSSL cput

odemengeon:
start of implementation

### DIFF
--- a/.github/tlspuffin.matrix.json
+++ b/.github/tlspuffin.matrix.json
@@ -175,8 +175,8 @@
       "preset": "wolfssl510"
     },
     "cargo": {
-      "name": "wolfssl510",
-      "features": []
+      "name": "",
+      "features": ["cputs"]
     }
   },
   {
@@ -187,8 +187,9 @@
       "preset": "wolfssl510-sig"
     },
     "cargo": {
-      "name": "wolfssl510",
+      "name": "",
       "features": [
+        "cputs",
         "fix-CVE-2022-25640",
         "fix-CVE-2022-39173"
       ]
@@ -202,8 +203,9 @@
       "preset": "wolfssl510-skip"
     },
     "cargo": {
-      "name": "wolfssl510",
+      "name": "",
       "features": [
+        "cputs",
         "fix-CVE-2022-25638",
         "fix-CVE-2022-39173"
       ]
@@ -217,8 +219,8 @@
       "preset": "wolfssl520"
     },
     "cargo": {
-      "name": "wolfssl520",
-      "features": []
+      "name": "",
+      "features": ["cputs"]
     }
   },
   {
@@ -229,9 +231,9 @@
       "preset": "wolfssl520-asan"
     },
     "cargo": {
-      "name": "wolfssl520",
+      "name": "",
       "features": [
-        "asan"
+        "cputs"
       ]
     }
   },
@@ -245,8 +247,8 @@
       "preset": "wolfssl530"
     },
     "cargo": {
-      "name": "wolfssl530",
-      "features": []
+      "name": "",
+      "features": ["cputs"]
     }
   },
   {
@@ -259,9 +261,9 @@
       "preset": "wolfssl530-asan"
     },
     "cargo": {
-      "name": "wolfssl530",
+      "name": "",
       "features": [
-        "asan"
+        "cputs"
       ]
     }
   },
@@ -275,8 +277,8 @@
       "preset": "wolfssl540"
     },
     "cargo": {
-      "name": "wolfssl540",
-      "features": []
+      "name": "",
+      "features": ["cputs"]
     }
   },
   {
@@ -287,8 +289,9 @@
       "preset": "wolfssl540-sdos2"
     },
     "cargo": {
-      "name": "wolfssl540",
+      "name": "",
       "features": [
+        "cputs",
         "wolfssl-disable-postauth",
         "fix-CVE-2022-39173"
       ]
@@ -302,8 +305,9 @@
       "preset": "wolfssl530-cdos"
     },
     "cargo": {
-      "name": "wolfssl530",
+      "name": "",
       "features": [
+        "cputs",
         "fix-CVE-2022-39173"
       ]
     }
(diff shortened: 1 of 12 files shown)


LCBH:
I started to comment the PR but it seems that your rebase removed some recent changes from other PRs.

client_log_config.yml
```diff
@@ -1,113 +0,0 @@
-# Scan this file for changes every 30 seconds
```
LCBH, line 1: Why removing this file?

.github/tlspuffin.matrix.json
```diff
@@ -259,9 +261,9 @@
       "preset": "wolfssl530-asan"
     },
     "cargo": {
-      "name": "wolfssl530",
+      "name": "",
```
LCBH, line 264: Please double check that the same CI tests get triggered (now with CPUT instead of Rust PUT); i.e., make sure the tests filtering nehave the same for CPUT and Rust PUT.

puffin-build/vendors/wolfssl/presets.toml
```diff
@@ -123,4 +123,4 @@ sancov = true
 [wolfssl430]
 sources = { repo = "https://github.com/wolfSSL/wolfssl.git", branch = "v4.3.0-stable", version = "4.3.0" }
 builder = { type = "builtin", name = "wolfssl" }
-sancov = true
\ No newline at end of file
+sancov = true
```
LCBH, line 126: What's here?

tlspuffin/tests/determinism.rs
```diff
@@ -1,6 +1,10 @@
+use std::thread;
+use std::time::Duration;
+
 use tlspuffin::test_utils::prelude::*;
 
-#[apply(test_puts, filter = all(tls13, boringssl))]
+#[cfg(not(feature = "wolfssl430"))]
+#[apply(test_puts, filter = all(tls13, any(boringssl, wolfssl)))]
```
LCBH, line 7: - Replace filtering by `not(openssl)` (better when we'll add new PUTs in teh future).
 - Please add a comment explaining why we currently exclude openssl.

tlspuffin/tests/vulnerabilities.rs
```diff
@@ -120,8 +120,9 @@ fn test_seed_cve_2022_38153(put: &str) {
     let runner = default_runner_for(put);
     let trace = seed_successful12_with_tickets.build_trace();
 
-    for _ in 0..50 {
-        let _ = runner.execute(trace.clone()).unwrap();
+    let _ = runner.execute_config(trace.clone(), true).unwrap();
```
LCBH, line 123: Please add a comment here:
```
/* 
Originally, puffin found this bug because wolfssl was not made deterministic at all. The bug requires that the
shared (across sessions) ticket map gets filled until a collision happen (a key refers to two tickets). For this to
happen, we need to create several **different** tickets. This won't happen, now that wolfssl is made
deterministic (the same ticket will be created at the end of a single handshake). We mimick the old behavior
here, using `.execute_config`.

Theoretically, this attack can still be found with a deterministic WolfSSL: mutations could repeat a full
handshake multiple times (hence yielding different tickets) and then appending the malicious handshake.
This is extremely unlikely to happen though; or even impossible given the bounds on the trace lengths. In
the future, we might want to reconsider whether we **always** want to reseed prior to executing a trace.
*/

```

tools/asanalyzer.py
```diff
@@ -129,4 +129,4 @@ def main():
             print("\t\t%s" % dup)
 
 if __name__=='__main__':
-    main()
\ No newline at end of file
+    main()
```
LCBH, line 132: What's here?

LCBH (changes requested):

LCBH:
Please close the old, outdated comments and resolve the new ones. We can merge after that.

tlspuffin-claims/claim-interface.h
```diff
@@ -1,9 +1,12 @@
 #ifndef TLSPUFFIN_CLAIM_INTERFACE_H
 #define TLSPUFFIN_CLAIM_INTERFACE_H
 
+#include <stdint.h>
+
 #define CLAIM_MAX_AVAILABLE_CIPHERS 256
 #define CLAIM_MAX_SECRET_SIZE 64 /* longest known is SHA512 */
 #define CLAIM_SESSION_ID_LENGTH 32
+#define CLAIM_MAX_CERTIFICATE_LENGHT 10240
```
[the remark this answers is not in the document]
LCBH: Yes!

tlspuffin/harness/wolfssl/src/put.c
```diff
@@ -0,0 +1,1305 @@
...
+            *result_code = RESULT_ERROR_OTHER;
+            break;
+        }
+    }
+
+    if (error_code == WOLFSSL_ERROR_NONE)
+    {
+        return strdup("no wolfssl error");
+    }
+    char *error_msg = (char *)calloc(1, 81);
```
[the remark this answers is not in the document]
LCBH: @aeyno Are you sure your error example originates from `wolfSSL_ERR_error_string_n`? (Because it is true that its output is capped by WOLFSSL_MAX_ERROR_SZ (80 by default)).

tlspuffin/tests/determinism.rs
LCBH: Comment to add:
```rust
/*
OpenSSL is currently excuded because: (i) we currently don't have a working RNG reseed for the OpenSSL CPUT (see Issue #393), (ii) we don't have a time hook at the moment for the OpenSSL CPUT.
*/
```

tlspuffin/harness/wolfssl/src/put.c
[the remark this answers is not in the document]
LCBH: After investigation: Rust PUT computes a stack of error codes here https://github.com/tlspuffin/tlspuffin/blob/84cbf310e042e773472747ce0b1f0bba5e7ead9f/crates/wolfssl/src/error.rs#L19

LCBH (approved):

PR #385, opened 2025-04-10
Allow setting ciphers for tls12

aeyno:
This PR introduce two separates API to set TLS 1.2 and 1.3 ciphersuites used in PUTs. The PR also introduces unit tests for those APIs and capabilities `allow_setting_tls12_ciphers` and `allow_setting_tls13_ciphers` to test the PUTs with working APIs.


tlspuffin/harness/openssl/src/put.c
```diff
@@ -308,9 +313,14 @@ AGENT openssl_create_server(const TLS_AGENT_DESCRIPTOR *descriptor)
     SSL_CTX_set_tmp_rsa(ssl_ctx, rsa);
 #endif
 
-    // Allow EXPORT in server
-    SSL_CTX_set_cipher_list(ssl_ctx, descriptor->cipher_string);
-    SSL_CTX_set_ciphersuites(ssl_ctx, descriptor->cipher_string);
+    if (descriptor->tls_version == V1_3)
+    {
+        SSL_CTX_set_ciphersuites(ssl_ctx, descriptor->cipher_string_tls13);
```
LCBH, line 318: Don't you want to first check that the `allow_setting_tls12/3_ciphers` capability is available?

tlspuffin/src/protocol.rs
```diff
@@ -339,7 +341,8 @@ impl Default for TLSDescriptorConfig {
             server_authentication: true,
             try_reuse: false,
             typ: AgentType::Server,
-            cipher_string: String::from("ALL:!EXPORT:!LOW:!aNULL:!eNULL:!SSLv2"),
+            cipher_string_tls13: String::from("ALL:!EXPORT:!LOW:!aNULL:!eNULL:!SSLv2"),
```
LCBH, line 344: Suggestion: use a global const string instead of inlining here

tlspuffin/src/rust_put/openssl/mod.rs
```diff
@@ -218,8 +240,27 @@ impl RustPut {
...
-        ctx_builder.set_cipher_list("ALL:!EXPORT:!LOW:!aNULL:!eNULL:!SSLv2")?;
+        match descriptor.protocol_config.tls_version {
+            TLSVersion::V1_3 => {
+                #[cfg(any(
+                    feature = "openssl101-binding",
+                    feature = "openssl102-binding",
+                    feature = "openssl111-binding",
+                    feature = "libressl-binding"
+                ))]
+                ctx_builder.set_cipher_list(&descriptor.protocol_config.cipher_string_tls13)?;
```
LCBH, line 251: please add a comment explaining tha the API changed for those versions, if that is the case

tlspuffin/src/tls/seeds.rs
```diff
@@ -397,6 +397,51 @@ pub fn seed_successful12(client: AgentName, server: AgentName) -> Trace<TLSProto
     }
 }
 
+pub fn seed_successful12_forward(client: AgentName, server: AgentName) -> Trace<TLSProtocolTypes> {
```
LCBH, line 400: Add a comment explaining what this should do

LCBH (approved):
LGTM, except minor issues

tlspuffin/harness/openssl/src/put.c
aeyno: Capabilities are not accessible in this function, I don't really want to overhaul the whole c put architecture for this check. With the default TLS string the PUTs work normally but if somebody tries to set a different cipher on a PUT that don't support it, the PUT will already throw an error.

tlspuffin/src/protocol.rs
aeyno: Done

tlspuffin/src/rust_put/openssl/mod.rs
aeyno: Done

tlspuffin/src/tls/seeds.rs
aeyno: Done

PR #386, opened 2025-04-17
Introduce display-execute cli command

aeyno:


puffin/src/trace.rs
```diff
@@ -337,6 +343,8 @@ pub struct TraceContext<PB: ProtocolBehavior> {
     spawner: Spawner<PB>,
 
     phantom: PhantomData<PB>,
+    /// The number of steps that have been successfully executed
+    pub executed_until: usize,
 }
```
LCBH, line 348: hmm, this is weird. A trace context is not necessarily bound to a single trace.

puffin/src/trace.rs
```diff
@@ -485,6 +494,255 @@ pub struct Trace<PT: ProtocolTypes> {
     pub prior_traces: Vec<Trace<PT>>,
 }
 
+/// Store the result of a trace execution for displaying or serializing
+#[derive(Serialize)]
+#[serde(bound = "PT: ProtocolTypes")]
+pub struct ExecutionResult<PT: ProtocolTypes> {
+    put: String,
+    success: bool,
+    error: Option<String>,
```
LCBH, line 503: Is `success` required here? Or is it equivalent to have `error = None`?

puffin/src/trace.rs
```diff
@@ -485,6 +494,255 @@ pub struct Trace<PT: ProtocolTypes> {
...
+        match self.success {
+            true => writeln!(f, "Success"),
+            false => writeln!(f, "Error : {}", self.error.clone().unwrap_or("".into())),
+        }
+    }
+}
+
+#[derive(Serialize)]
+#[serde(bound = "PT: ProtocolTypes")]
+pub struct TraceExecution<PT: ProtocolTypes> {
```
LCBH, line 553: Add a Rust Doc

LCBH (changes requested):

puffin/src/trace.rs
aeyno: It is equivalent, it is fixed

puffin/src/trace.rs
aeyno: done

puffin/src/trace.rs
aeyno: Ok, I see `executed_until` should be the number of step executed in the last trace (excluding prior traces). As we discussed, recursive prior traces should be removed in favor of a Vec of traces, but at the moment counting the number of executed steps with potentially recursive prior traces is too much work for no real benefit

aeyno: is it ok ?

puffin/src/trace.rs
```diff
@@ -680,7 +942,12 @@ impl<PT: ProtocolTypes> OutputAction<PT> {
         }
     }
 
-    fn execute<PB>(&self, agent_name: AgentName, ctx: &mut TraceContext<PB>) -> Result<(), Error>
+    fn execute<PB>(
```
LCBH, line 945: Please document the step argument. It was unclear to me at first sight.

LCBH (changes requested):

puffin/src/trace.rs
aeyno: done

aeyno:
Can we merge now ?

LCBH (approved):

PR #392, opened 2025-05-12
Fix OpenSSL not setting correctly the cipher list 

aeyno:
OpenSSL doesn't behave as described in its documentation (https://docs.openssl.org/3.1/man3/SSL_CTX_set_cipher_list/) : to restrict the list of available ciphers in TLS 1.3, using only `SSL_CTX_set_ciphersuites` is not sufficient (it only changes the order of the ciphers). To correctly restrict the available ciphers, only `SSL_CTX_set_cipher_list` (which should be used for TLS 1.2 and below) works.
This PR uses both `SSL_CTX_set_cipher_list` and  `SSL_CTX_set_ciphersuites` to set the correct cipher list in the correct order for OpenSSL


LCBH (approved):
LGTM

PR #411, opened 2025-07-18
[Bit-4] New Error::Codec and more specific error types in key exchange and codec functions

LCBH:
refactor: update error handling (and with new Error::Codec) to use more specific error types in key exchange and codec functions


aeyno (approved):
LGTM

PR #412, opened 2025-07-18
[Bit-5] refactor: with-bit now opt-in + refactor logging

LCBH:
refactor: with-bit is now opt-in + enhance logging levels and configurations for improved traceability and debugging
Add a `--verbosity <LevelFilter>` cli command to change the debugging level during quick experiment.

### DIFF
--- a/tlspuffin/src/tls/rustls/msgs/handshake_test.rs
+++ b/tlspuffin/src/tls/rustls/msgs/handshake_test.rs
@@ -462,7 +462,7 @@ fn test_truncated_client_extension_is_detected() {
 
     for ext in &chp.extensions.0 {
         let mut enc = ext.get_encoding();
-        // println!("testing {:?} enc {:?}", ext, enc);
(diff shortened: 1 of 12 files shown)


tlspuffin/src/tls/rustls/msgs/handshake_test.rs
```diff
@@ -462,7 +462,7 @@ fn test_truncated_client_extension_is_detected() {
 
     for ext in &chp.extensions.0 {
         let mut enc = ext.get_encoding();
-        // println!("testing {:?} enc {:?}", ext, enc);
+        // log::trace!("testing {:?} enc {:?}", ext, enc);
```
aeyno, line 465: should we keep this comment ?

aeyno (approved):

tlspuffin/src/tls/rustls/msgs/handshake_test.rs
LCBH: I kept it to make sure one can resurrect logging in the future should one want to.

PR #417, opened 2025-07-21
[Bit-7] Pr/bit mutations/rework term zoo

LCBH:
Should be thoroughly tested since this could impact `GenerateMutator`.

[fuzzer: rework TermZoo.](https://github.com/tlspuffin/tlspuffin/pull/417/commits/00f7abe2b7425a536a42f5509e860a617638244d)
 - allow to filter out generated terms that do not successfully evaluate: `filter_evaluated`

 - use the `[no_gen]` signature attribute to filter out symbols for which we won't be able to generate terms

 - GenerateMutator makes use of all of that to generate potentially more useful terms

 - Adapt uni tests (term_zoo) and benchmarks (compute_zoo_in_generate_mutator)

 - Also added `test_term_payloads_mutate_eval` checking the function symbols for which we can generate a term, MakeMessage, BitFlipMutator, successfully evaluate. Corresponding benchmark: benchmark_test_term_payloads_mutate_eval

### DIFF
--- a/puffin/src/fuzzer/libafl_setup.rs
+++ b/puffin/src/fuzzer/libafl_setup.rs
@@ -11,7 +11,7 @@ use log4rs::Handle;
 
 use super::harness;
 use crate::fuzzer::feedback::MinimizingFeedback;
-use crate::fuzzer::mutations::{trace_mutations, MutationConfig};
+use crate::fuzzer::mutations::{dy_mutations, MutationConfig};
 use crate::fuzzer::stages::PuffinMutationalStage;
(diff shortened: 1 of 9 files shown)


aeyno (commented):
I'm not fully aware of how all the term_zoo stuff works, so I reviewed this PR but I'm not sure that it was really helpful

LCBH:
> I'm not fully aware of how all the term_zoo stuff works, so I reviewed this PR but I'm not sure that it was really helpful

No worries. We will be careful evaluating the potential impact on performance though since zoo generation is likely to require more ressources.

PR #418, opened 2025-07-21
[Bit-8] Pr/bit-mutations/change-tls-sig

LCBH:
Needs to be thoroughly evaluated due to various modifications to the TLS signature.

 - [fix-tls: refactor certificate entry functions to include extensions and improve chaining](https://github.com/tlspuffin/tlspuffin/pull/418/commits/7d4173ea602536365c3fde6ed0b00c17a6a93466)
   + consistent argument ordering for list symbols and use of wrapped item
   + make sure append function symbols append in the order of the arguments (otherwise, right sibling payload look up and replacement won't work)
- [fuzzing-tls: remove "size cheating" symbols from TLS signature: remove size-related custom values such as fake cred and fake heartbeat sizes (should be reproduced with bit-level mutations now)](https://github.com/tlspuffin/tlspuffin/pull/418/commits/8346eafdcb1f89beb68b5d2e137b80dd365e4467)
- [fuzzing-tls: explicit u32 to u16 coercion function symbol](https://github.com/tlspuffin/tlspuffin/pull/418/commits/a243845e014f378f802bf7753732c966343fa06a)

### DIFF
--- a/tlspuffin/src/tls/fn_extensions.rs
+++ b/tlspuffin/src/tls/fn_extensions.rs
@@ -106,6 +106,12 @@ pub fn fn_cert_extensions_append(
     Ok(new_extensions)
 }
 
+pub fn fn_cert_extensions_make(
+    extensions: &Vec<CertificateExtension>,
+) -> Result<CertificateExtensions, FnError> {
+    Ok(CertificateExtensions(extensions.clone()))
+}
+
 pub fn fn_new_session_ticket_extensions_new() -> Result<Vec<NewSessionTicketExtension>, FnError> {
     Ok(vec![])
 }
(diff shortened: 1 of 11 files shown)


tlspuffin/src/tls/fn_extensions.rs
```diff
@@ -636,41 +642,41 @@ pub fn fn_transport_parameters_draft_server_extension(
...
-        payload: Payload::new([42; 7000]),
+        payload: Payload::new([42; 10]),
     }))
 }
 
 pub fn fn_unknown_server_extension() -> Result<ServerExtension, FnError> {
     Ok(ServerExtension::Unknown(UnknownExtension {
         typ: ExtensionType::Unknown(0xFFFF),
-        payload: Payload::new([42; 7000]),
+        payload: Payload::new([42; 10]),
```
aeyno, line 652: Are there any reason to use 10 as payload size and why was it 7000 before ? I suppose that 7000 was impacting the amount of bitlevel modifications that could happen is such a large payload.

aeyno (approved):
LGTM

tlspuffin/src/tls/fn_extensions.rs
LCBH: The rationale is that we will have bit-level mutations to extend such vectors if needed (to find something interesting). Conversely, always use very large vectors create spurious data on which bit-level mutations will operate (relevance thereof is questionable).

PR #419, opened 2025-08-11
Add symbols to build list of supported groups

aeyno:
This PR adds `fn_support_group_extension_new`, `fn_support_group_extension_make` and `fn_support_group_extension_append` function symbols to create lists of supported groups in ClientHello. This PR also removes `fn_support_group_extension` which can be replaced by the new symbols.

### DIFF
--- a/tlspuffin/tests/term_zoo.rs
+++ b/tlspuffin/tests/term_zoo.rs
@@ -214,7 +214,7 @@ fn test_term_dy_eval() {
         let res = zoo_test(
             |term, ctx, _| term.evaluate_dy(&ctx).map(|_| ()),
             rand,
-            2600,
+            3700,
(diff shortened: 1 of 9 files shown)


tlspuffin/tests/term_zoo.rs
```diff
@@ -214,7 +214,7 @@ fn test_term_dy_eval() {
         let res = zoo_test(
             |term, ctx, _| term.evaluate_dy(&ctx).map(|_| ()),
             rand,
-            2600,
+            3700,
```
LCBH, line 217: Is this the mnimal value for which the tests pass?

LCBH (changes requested):

LCBH (approved):

PR #420, opened 2025-08-19
Add a TLS 1.3 seed with a HelloRetryRequest

aeyno:
This PR introduces `seed_server_attacker_with_hello_retry_request` which trigger a TLS implementation to send a second ClientHello with a different keyshare, along with the necessary function symbol to compute the transcript hash and some patches to the encoding of HelloRetryRequest payload type.

Note that HelloRetryRequest in tlspuffin are currently encoded as ServerHello messages and that we currently could not send a real HelloRetryRequest message to the implementations. This is not a real problem since all the tested implementations (OpenSSL, WolfSSL and BoringSSL) reject HelloRetryRequest messages and only accept ServerHello as HRR.

### DIFF
--- a/tlspuffin/src/tls/rustls/msgs/handshake.rs
+++ b/tlspuffin/src/tls/rustls/msgs/handshake.rs
@@ -1211,8 +1211,15 @@ impl codec::Codec for HelloRetryRequest {
         self.random.encode(bytes);
         self.session_id.encode(bytes);
         self.cipher_suite.encode(bytes);
-        // self.compression_methods.encode(bytes);
-        Compression::Null.encode(bytes);
+        // HelloRetryRequest is encoded as a ServerHello
+        // ServerHello only contains one compression method while HRR can hold a vec of compression
+        // methods if the vec is not empty we encode the first element, else we encode the
+        // Null compression
+        if !self.compression_methods.0.is_empty() {
+            self.compression_methods.0[0].encode(bytes);
+        } else {
+            Compression::Null.encode(bytes);
+        }
         self.extensions.encode(bytes);
     }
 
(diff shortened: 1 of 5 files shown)


tlspuffin/src/tls/rustls/msgs/handshake.rs
```diff
@@ -1211,20 +1211,17 @@ impl codec::Codec for HelloRetryRequest {
         self.random.encode(bytes);
         self.session_id.encode(bytes);
         self.cipher_suite.encode(bytes);
-        self.compression_methods.encode(bytes);
-        // Compression::Null.encode(bytes);
-        // if !self.extensions.0.is_empty() {  // TODO: @MAX shouldn't we also accept empty ones?
-        //         // lists of extensions, as for CLientHello ClientExtensions?
+        // self.compression_methods.encode(bytes);
+        Compression::Null.encode(bytes);
```
LCBH: Take first if any, Null otherwise

tlspuffin/src/tls/seeds.rs
```diff
@@ -781,6 +781,226 @@ pub fn seed_server_attacker_full(client: AgentName) -> Trace<TLSProtocolTypes> {
     }
 }
 
+pub fn seed_server_attacker_with_hello_retry_request(client: AgentName) -> Trace<TLSProtocolTypes> {
```
LCBH, line 789: Add doc plz

LCBH (changes requested):

tlspuffin/src/tls/seeds.rs
aeyno: done

tlspuffin/src/tls/rustls/msgs/handshake.rs
aeyno: done

LCBH (approved):

PR #423, opened 2025-08-21
[Bit-11] Bit-level mutations optimizations

LCBH:
- [fuzzer: enforce payload-to-terms ratio constraints in ReplaceReuseMutator and MakeMessage](https://github.com/tlspuffin/tlspuffin/pull/423/commits/d60d7126702c856ba1a1a0ddfefc263563f941e0)
   + Avoid exceeding the payload-per-term threshold by skipping MakeMessage application when needed.
   + Avoid exceeding the payload-per-term threshold by skipping ReplaceMutator when the chosen replacement would yield too many payloads in the resulting trace.ped mutations due to excessive payloads.
 - [fuzzing: when MakeMessage on a non fully symbolic term, payload_0:=evaluate_symbolic() but payload=evaluate(), hence keeping benefits from previous bit-level mutations on those](https://github.com/tlspuffin/tlspuffin/pull/423/commits/e6d9e3d5d8e32e2158fba99772206b92d4f0c0b7)

### DIFF
--- a/puffin/src/fuzzer/mutations.rs
+++ b/puffin/src/fuzzer/mutations.rs
@@ -45,11 +45,11 @@ impl Default for MutationConfig {
 
 pub type DyMutations<'harness, PT, PB, S> = tuple_list_type!(
 // DY mutations
-      RepeatMutator<S>,
-      SkipMutator<S>,
-      ReplaceReuseMutator<S>,
-      ReplaceMatchMutator<S, PT>,
-      RemoveAndLiftMutator<S>,
+    RepeatMutator<S>,
+    SkipMutator<S>,
+    ReplaceReuseMutator<S>,
+    ReplaceMatchMutator<S, PT>,
+    RemoveAndLiftMutator<S>,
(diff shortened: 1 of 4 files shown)


puffin/src/fuzzer/mutations.rs
```diff
@@ -423,13 +423,29 @@ where
             return Ok(MutationResult::Skipped);
         }
         let rand = state.rand_mut();
+        let (trace_nb_payloads, nb_terms) = if self.with_bit {
+            (trace.all_payloads().len(), trace.steps.len())
```
aeyno, line 427: Not sure why `trace.steps.len()` corresponds to `nb_terms?  A trace can be composed of only output steps without any terms

aeyno (changes requested):
LGTM

puffin/src/fuzzer/mutations.rs
LCBH: Indeed, I will fix that in future PR and linked to #444

aeyno (approved):

PR #431, opened 2025-09-25
Force reading PUT buffer after output even at failing step

aeyno:
This PR allows reading the content of the PUT output buffer during an input step triggering an error. Changes :
- During the input step, if the PUT returns an error we first trigger an output buffer read before returning the error and stopping the trace execution
- `take_message_from_outbound` now takes a mutable reference to a `MessageFlight` that will be filled with the messages on the buffer before stopping when an error is returned instead of returning nothing if there is an error

### DIFF
--- a/puffin/src/trace.rs
+++ b/puffin/src/trace.rs
@@ -1071,7 +1071,7 @@ impl<PT: ProtocolTypes> Step<PT> {
                 let _ = (OutputAction {
                     phantom: Default::default(),
                 })
-                .execute(self.agent, step_number, ctx);
+                .read_state(self.agent, step_number, ctx);
                 exec
             }
             Action::Output(output) => output.execute(self.agent, step_number, ctx),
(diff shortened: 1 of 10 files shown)


puffin/src/trace.rs
```diff
@@ -1014,13 +1014,15 @@ impl<PT: ProtocolTypes> Step<PT> {
...
+            Action::Input(input) => {
+                let exec = input.execute(self.agent, ctx);
                 // NOTE force output after each InputAction step
-                (OutputAction {
+                let _ = (OutputAction {
                     phantom: Default::default(),
                 })
-                .execute(self.agent, step_number, ctx)
-            }),
+                .read_state(self.agent, step_number, ctx);
```
LCBH, line 1074: Why not `execute`?

LCBH: Got it, `read_state` what we want here.

tlspuffin/src/put.rs
```diff
@@ -362,13 +364,18 @@ impl Stream<TLSProtocolBehavior> for CAgent {
...
+                            *output_flight = (!flight.messages.is_empty()).then_some(flight);
+                            return Err(err.into());
+                        }
                     },
                 }
             }
         }
 
-        Ok((!flight.messages.is_empty()).then_some(flight))
+        *output_flight = (!flight.messages.is_empty()).then_some(flight);
```
LCBH, line 376: So we cannot return `Some(vec![])`, is this intended?

LCBH (approved):

tlspuffin/src/put.rs
LCBH: TODO: remove Option type

aeyno: After a bit more thought, I think that keeping the Option is be a better idea, especially because this function might be called after Input steps that would trigger no output (for example TLS ServerHello where the PUT will wait for encrypted messages) but create knowledges with empty flight. This would result in the number of flights in knowledges matching the number of steps in the trace rather than the number of flights sent by a PUT.

LCBH: OK I get it. Do you mean you would keep a None in output_flight if this function fails to push any new message to `flight`?

LCBH (approved):

tlspuffin/src/put.rs
aeyno: Yes, None means the output buffer of the PUT was empty so no flight was received

aeyno: Yes, None means the output buffer of the PUT was empty so no flight was received

PR #436, opened 2025-11-13
fix a bug introduced in PR 330 that disabled --put-use-clear

LCBH:
fix PR #330: [A bug](https://github.com/tlspuffin/tlspuffin/pull/330/commits/175a6b578f7a9c3dc8ee9b1c02f1a235c4453389#r2524251103) in refactoring spawner was defaulting AgentDescriptor to static default instead of using configuration given from the CLI. Fixed with a new `default_put` field in `PutRegistry`. Sets it from the CLI, reads it before Spawn.

Effect: E2E testing for reproducing SDOS2 was broken. We could find it through uni testing (that uses the right agent configuration with use_clear) but not through fuzzing anymore since the option --put-use-clear has no effect.

This fixes the problem, SDOS2 is found again in seconds.

Replaced redundant use of `PutDescriptor` with a unified `default_put` in `PutRegistry`. Introduced `PutOptions` to manage default PUT options centrally and simplified related code paths accordingly.

### DIFF
--- a/puffin/src/algebra/mod.rs
+++ b/puffin/src/algebra/mod.rs
@@ -665,6 +665,7 @@ mod tests {
     use crate::algebra::signature::Signature;
     use crate::algebra::term::TermType;
     use crate::algebra::{AnyMatcher, DYTerm, Term};
+    use crate::put::{PutDescriptor, PutOptions};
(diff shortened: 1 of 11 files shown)


aeyno (approved):
LGTM

Issue #414, opened 2025-07-21
Improve mk_vendor error handling


LCBH:
Fix non-informative error message in case of typos such as:
```
./tools/mk_vendor make openss:openssl340-asan
thread 'main' panicked at /home/lhirschi/tlspuffin_byte/puffin-build/src/library/build/config.rs:46:10:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
./tools/mk_vendor : ligne 48 : 2272239 Aborted                 "${TARGET_DIR}/debug/mk_vendor" "$@
```

PR #439, opened 2025-12-04
Fix: handle mk_vendor panic when vendor is not found

tathanhdinh:
Hello all,

This small pull request tries to fix the problem raised in #414.
Now if the vendor or preset are not valid, the `mk_vendor` tool will show error messages instead of panic. E.g.

```
./tools/mk_vendor make openss:openssl312-asan
[2025-12-04T16:29:15Z ERROR mk_vendor] configuration vendor 'openss' and (or) preset 'openssl312-asan' not found
```

Many thanks for any comment.


LCBH:
Thank you @tathanhdinh! The changes look good to me and the new error message is now sufficiently precise. Will merge soon (we need to finish up other merges first).

aeyno (approved):

PR #457, opened 2026-02-11
Libafl update to 0.15.4

nataelbaffou:
Updates LibAFL from 0.12.0 to 0.15.4

## Dependency update
- crates/boring
  - `bitflags` : 2.6 to 2.10.0
- crates/wolfssl
  - `bitflags` : 2.6 to 2.10.0
- puffin
  - `libafl` : 0.12.0 to 0.15.4
  - `libafl_targets` : 0.12.0 to 0.15.4
  - `libafl_bolts` : 0.12.0 to 0.15.4
- tlspuffin
  - fixed `openssl-sys` to 0.9.85 (lots of breaking changes in 0.9.111 current latest)
  - fixed `openssl-src` to 111.22.0
- project
  - fixed `comparable` to 0.5.5 (0.5.6 no longer supports arrays)

Updated cargo.lock

## Changes

The following changes are from the crate changes itself, when preceded by [PUF] then it is a "puffin" change even so they were modified because of the update, they are not a type or trait from the LibAFL crate

Lot's of breaking changes from libafl update: (list non exhaustive of changes made)

- BytesInput methods from `bytes()` to `mutator_bytes()` and `bytes_mut()` to `mutator_bytes_mut()`
- UsesInput and UsesState removed from existence -> have to use generics instead
- `PT` generic type added in many bit_mutation mutators and functions
- `HasCorpus` trait needs generic I, often `Trace<PT>`
- `post_exec()` method needed for types `Mutator`, 
- `Named` trait `name` method now returns &Cow<'static, str> -> forces to use static names
- `Feedback` needs `StateInitializer` trait
- `Feedback` from 1 to 4 generics (EM, I, OT, SC)
- `Feedback` methods `is_interesting`, `is_interesting_introspection` and `append_metadata` takes 2 generics (EM, OT)
- `Input` trait `generate_name` method `_idx` parameter type from `usize` to `Option<CorpusId>`
- `StdState` generics order changed from `<I, C, R, SC>` to `<C, I, R, SC>`
- `ComposedByMutations` trait from 3 to 0 generics
- `ComposedByMutations` added attribute `type Mutations`
- `ScheduledMutator` trait from 3 `<I, MT, S>` to 2 generics `<I, S>`
- `TuiUI` renamed to `TuiUi`
- `TuiMonitor` now uses a builder instead of new
- `Monitor` no longer has stats directly -> now stored in ClientStatsManager
- `Stage` trait from 3 `<E, EM, Z>` to 4 `<E, EM, S, Z>` generics
- `StagesTuple` needs `Stage` to be `Restartable`
- [PUF] `StatsStage` struct from 3 `<E, EM, Z>` to 5 `<E, EM, Z, S, I>` generics
- [PUF] `StatsStage` added `name` attribute and `Named` trait
- `Evaluator` trait from 2 `<E, EM` to 4 `<E, EM, I, S>` generics
- `Event` firing changed. Fires `EventWithStats`
- `InProcessExecutor` type from 3 `<H, OT, S>` to 6 `<EM, H, I, OT, S, Z>` generics
- [PUF] `ConcreteExecutor` type from 3 `<H, OT, S>` to 6 `<EM, H, I, OT, S, Z>` generics
- `EventManager` trait removed -> now every trait has to be specified (EventFirer, EventRestarter, EventReceiver...)
- `StdScheduledMutator` renamed to `HavocScheduledMutator`
- [PUF] removed `PuffinMutationalStage` and replaced by `StdMutationalStage`
- `StdFuzzer` struct from 4 `<CS, F, OF, OT>` to `<CS, F, IC, IF, OF>`
- renamed global `MAX_EDGES_NUM` to `MAX_EDGES_FOUND`
- `CombinedFeedback` struct from 4 `<A, B, FL, S>` to 3 `<A, B, FL>` generics (A and B are Feedbacks)
- [PUF] `ConcreteFeedback` type from 1 `<S>` to 0 generics
- `Launcher` builder method `run_client` changed its third parameter from `CoreId` to `ClientDescription`
- Libafl_bolts `Rand` struct `below` method is exclusive and no longer accepts 0 values. `below_or_zero` method created
- Libafl_bolts `Rand` struct `between` method changes parameter types from u64 to usize
- `ExplicitTracking` struct wrap over `Observers` for them to have tracking data
- `TimeoutExecutor` no longer exists as it is implemented in every executor
- [0.11.2 -> 0.12.0] `Mutators` trait `mutate` method no longer has `stage_idx` parameter
- [0.11.2 -> 0.12.0] `ScheduledMutators` trait `scheduled_mutate` method no longer has `stage_idx` parameter

### Other change references
- LibAFL releases: [link](https://github.com/AFLplusplus/LibAFL/releases)
- LibAFL [Migration.md](https://github.com/AFLplusplus/LibAFL/blob/main/MIGRATION.md) only for recent versions

### Generics names glossary (not official)
- EM: EventManager (often EventFirer)
- OT: ObserversTuple
- I: Input
- SC: state corpus (often InMemoryCorpus)
- S: state (often composition of traits like : HasExecution, HasRand, HasExecution, HasMetadata...)
- MT: MutatorsTuple
- E: Executor
- Z: Evaluator
- H: Harness function
- CS: Scheduler
- F: Feedback
- OF: Feedback ...
- IF: InputFilter
- IC: Converter
- FL: FeedbackLogic

## Fixes
- Wrongly randomized mutations in `mutations.rs/test_byte_remove_payloads`
- Test `test_term_payloads_mutate_eval` in `tlspuffin/tests/term_zoo.rs` failed with function symbol `fn_derive_psk` unsuccessful -> Increased the "how many" value

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,7 +77,7 @@ syn = "2.0.66"
 quote = "1.0.36"
 proc-macro2 = "1.0.79"
 trybuild = "< 1.0.91"
-comparable = "0.5.5"
+comparable = "= 0.5.5"
 
 [patch.crates-io]
 wolfssl-sys = { path = "crates/wolfssl-sys" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,11 @@ syn = "2.0.66"
 quote = "1.0.36"
 proc-macro2 = "1.0.79"
 trybuild = "< 1.0.91"
+
+# They changed the Comparable crate from 0.5.5 to 0.5.6 where arrays are no longer supported for now
+# 0.5.6 breaks tlspuffin/src/claims.rs::TlsTranscript array attribute
+# We can either use a vec<u8> or fix the current version of the Comparable crate
+# See related issue https://github.com/jwiegley/comparable/issues/13
 comparable = "= 0.5.5"
 
 [patch.crates-io]
--- a/puffin/src/fuzzer/bit_mutations.rs
+++ b/puffin/src/fuzzer/bit_mutations.rs
@@ -1,4 +1,5 @@
 use std::any::Any;
+use std::borrow::Cow;
 use std::cmp::min;
 use std::ops::Not;
 
@@ -54,70 +55,70 @@ use crate::trace::{ConfigTrace, Spawner, Trace, TraceContext};
     CrossoverReplaceMutator,
 */
 
-pub type HavocMutationsTypeDY<S> = tuple_list_type!(
-    BitFlipMutatorDY<S>,
-    ByteFlipMutatorDY<S>,
-    ByteIncMutatorDY<S>,
-    ByteDecMutatorDY<S>,
-    ByteNegMutatorDY<S>,
-    ByteRandMutatorDY<S>,
-    ByteAddMutatorDY<S>,
-    WordAddMutatorDY<S>,
-    DwordAddMutatorDY<S>,
-    QwordAddMutatorDY<S>,
-    ByteInterestingMutatorDY<S>,
-    WordInterestingMutatorDY<S>,
-    DwordInterestingMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesExpandMutatorDY<S>,
-    BytesLargeExpandMutatorDY<S>, // NEW! Different from classical havoc!!
-    BytesInsertMutatorDY<S>,
-    BytesRandInsertMutatorDY<S>,
-    BytesSetMutatorDY<S>,
-    BytesRandSetMutatorDY<S>,
-    BytesCopyMutatorDY<S>,
+pub type HavocMutationsTypeDY<S, PT> = tuple_list_type!(
+    BitFlipMutatorDY<S, PT>,
+    ByteFlipMutatorDY<S, PT>,
+    ByteIncMutatorDY<S, PT>,
+    ByteDecMutatorDY<S, PT>,
+    ByteNegMutatorDY<S, PT>,
+    ByteRandMutatorDY<S, PT>,
+    ByteAddMutatorDY<S, PT>,
+    WordAddMutatorDY<S, PT>,
+    DwordAddMutatorDY<S, PT>,
+    QwordAddMutatorDY<S, PT>,
+    ByteInterestingMutatorDY<S, PT>,
+    WordInterestingMutatorDY<S, PT>,
+    DwordInterestingMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesExpandMutatorDY<S, PT>,
+    BytesLargeExpandMutatorDY<S, PT>, // NEW! Different from classical havoc!!
+    BytesInsertMutatorDY<S, PT>,
+    BytesRandInsertMutatorDY<S, PT>,
+    BytesSetMutatorDY<S, PT>,
+    BytesRandSetMutatorDY<S, PT>,
+    BytesCopyMutatorDY<S, PT>,
     BytesInsertCopyMutatorDY<S>,
     BytesSwapMutatorDY<S>,
-    CrossoverInsertMutatorDY<S>,
-    CrossoverReplaceMutatorDY<S>,
-    SpliceMutatorDY<S>,
+    CrossoverInsertMutatorDY<S, PT>,
+    CrossoverReplaceMutatorDY<S, PT>,
+    SpliceMutatorDY<S, PT>,
 );
 
-pub type BitMutations<'harness, PB, S> = tuple_list_type!(
+pub type BitMutations<'harness, PB, PT, S> = tuple_list_type!(
     MakeMessage<'harness, PB>,
     ReadMessage<'harness, PB>,
-    BitFlipMutatorDY<S>,
-    ByteFlipMutatorDY<S>,
-    ByteIncMutatorDY<S>,
-    ByteDecMutatorDY<S>,
-    ByteNegMutatorDY<S>,
-    ByteRandMutatorDY<S>,
-    ByteAddMutatorDY<S>,
-    WordAddMutatorDY<S>,
-    DwordAddMutatorDY<S>,
-    QwordAddMutatorDY<S>,
-    ByteInterestingMutatorDY<S>,
-    WordInterestingMutatorDY<S>,
-    DwordInterestingMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesDeleteMutatorDY<S>,
-    BytesExpandMutatorDY<S>,
-    BytesLargeExpandMutatorDY<S>,
-    BytesInsertMutatorDY<S>,
-    BytesRandInsertMutatorDY<S>,
-    BytesSetMutatorDY<S>,
-    BytesRandSetMutatorDY<S>,
-    BytesCopyMutatorDY<S>,
+    BitFlipMutatorDY<S, PT>,
+    ByteFlipMutatorDY<S, PT>,
+    ByteIncMutatorDY<S, PT>,
+    ByteDecMutatorDY<S, PT>,
+    ByteNegMutatorDY<S, PT>,
+    ByteRandMutatorDY<S, PT>,
+    ByteAddMutatorDY<S, PT>,
+    WordAddMutatorDY<S, PT>,
+    DwordAddMutatorDY<S, PT>,
+    QwordAddMutatorDY<S, PT>,
+    ByteInterestingMutatorDY<S, PT>,
+    WordInterestingMutatorDY<S, PT>,
+    DwordInterestingMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesDeleteMutatorDY<S, PT>,
+    BytesExpandMutatorDY<S, PT>,
+    BytesLargeExpandMutatorDY<S, PT>,
+    BytesInsertMutatorDY<S, PT>,
+    BytesRandInsertMutatorDY<S, PT>,
+    BytesSetMutatorDY<S, PT>,
+    BytesRandSetMutatorDY<S, PT>,
+    BytesCopyMutatorDY<S, PT>,
     BytesInsertCopyMutatorDY<S>,
     BytesSwapMutatorDY<S>,
-    CrossoverInsertMutatorDY<S>,
-    CrossoverReplaceMutatorDY<S>,
-    SpliceMutatorDY<S>,
+    CrossoverInsertMutatorDY<S, PT>,
+    CrossoverReplaceMutatorDY<S, PT>,
+    SpliceMutatorDY<S, PT>,
 );
 
 pub type AllMutations<'harness, PT, PB, S> = tuple_list_type!(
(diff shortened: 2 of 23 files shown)


Cargo.toml
```diff
@@ -77,7 +77,7 @@ syn = "2.0.66"
 quote = "1.0.36"
 proc-macro2 = "1.0.79"
 trybuild = "< 1.0.91"
-comparable = "0.5.5"
+comparable = "= 0.5.5"
```
LCBH, line 85: Is this crucial?

puffin/src/fuzzer/mutations.rs
```diff
@@ -560,19 +574,23 @@ where
             log::debug!("       Skipped {}", self.name());
             return Ok(MutationResult::Skipped);
         }
-        let insert_index = state.rand_mut().between(0, length as u64) as usize;
-        let step = state.rand_mut().choose(steps).clone();
+        let insert_index = state.rand_mut().between(0, length);
+        let step = state.rand_mut().choose(steps);
```
LCBH: To avoid the potentially panicking unwrap below, use here `  let Some(step) = state.rand_mut().choose(steps) else {                                                                                                                               
      return Ok(MutationResult::Skipped);                                                                                                                                              
  };`

puffin/src/fuzzer/stats_stage.rs
```diff
@@ -368,24 +367,51 @@ where
...
-    E: UsesState<State = Z::State>,
-    EM: UsesState<State = Z::State>,
-    Z: Evaluator<E, EM>,
+    EM: EventFirer<I, S>,
+    E: Executor<EM, I, S, Z>,
+    Z: Evaluator<E, EM, I, S>,
+    I: Input,
 {
-    pub const fn new() -> Self {
+    pub fn new() -> Self {
```
LCBH, line 395: Use `        let stage_id = STATS_STAGE_ID.fetch_add(1, Ordering::Relaxed);` instead

puffin/src/fuzzer/stats_stage.rs
```diff
@@ -307,58 +311,53 @@ impl Fire for MinMaxMean {
     }
 }
 
+static mut STATS_STAGE_ID: usize = 0;
```
LCBH: Use `static STATS_STAGE_ID: AtomicUsize = AtomicUsize::new(0);` instead

puffin/src/fuzzer/stages.rs
```diff
@@ -43,59 +43,64 @@ where
...
 
-impl<I, MT, MtPre, MtPost, S> ComposedByMutations<I, MT, S>
-    for FocusScheduledMutator<I, MT, MtPre, MtPost, S>
+impl<I, MT, MtPre, MtPost, S> ComposedByMutations for FocusScheduledMutator<I, MT, MtPre, MtPost, S>
 where
     MT: MutatorsTuple<I, S>,
     MtPre: MutatorsTuple<I, S>,
     MtPost: MutatorsTuple<I, S>,
     S: HasRand,
 {
```
LCBH, line 79: Use the following instead:
```
    type Mutations = MT;

    /// Get the core mutations (pre/post are handled separately in scheduled_mutate)
    fn mutations(&self) -> &Self::Mutations {
        &self.mutations_core
    }

    /// Get the core mutations (mutable)
    fn mutations_mut(&mut self) -> &mut Self::Mutations {
        &mut self.mutations_core
    }
```

LCBH (changes requested):

tlspuffin/tests/term_zoo.rs
```diff
@@ -360,7 +359,7 @@ fn test_term_payloads_mutate_eval() {
         let res = zoo_test(
             &mut closure,
             StdRand::with_seed(i),
-            100,
```
LCBH, line 363: Why +200 since we don't change the signature in this PR?!

LCBH: If needed, change the seed (of randomness) instead.

tlspuffin/tests/mutations.rs
```diff
@@ -242,6 +243,11 @@ fn test_byte_remove_payloads(put: &str) {
             match &first.action {
                 Action::Input(input) => {
                     if let DYTerm::Application(_fd, args) = &input.recipe.term {
+                        if !input.recipe.is_symbolic() {
```
LCBH, line 248: Add a comment here: we revert to original trace if we applied a MakeMessage on the "wrong" sub-term.

tlspuffin/src/claims.rs
```diff
@@ -12,6 +12,9 @@ use smallvec::SmallVec;
 
 use crate::protocol::{AgentType, TLSProtocolTypes, TLSVersion};
 
+// They changed the Comparable crate from 0.5.5 to 0.5.6 where arrays are no longer supported for
+// now We can either use a vec<u8> or fix the current version of the Comparable crate (that's what I
```
LCBH: Please move this comment to where we fix the version (+ link to here in the cimment)

tlspuffin/benches/benchmark.rs
```diff
@@ -340,7 +340,7 @@ fn benchmark_test_term_payloads_mutate_eval(c: &mut Criterion) {
             let res = zoo_test(
                 &mut closure,
                 StdRand::with_seed(i),
-                100,
+                300,
```
LCBH, line 343: Same here: rather change the seed (of randomness) instead of increasing this.

puffin/src/fuzzer/stats_stage.rs
LCBH: I've seen this seems to be the way libAFL has implemented this. Let us discuss this point live.

puffin/src/fuzzer/stats_stage.rs
```diff
@@ -368,24 +367,51 @@ where
...
+impl<E, EM, S, Z, I> Restartable<S> for StatsStage<E, EM, S, Z, I>
+where
+    EM: EventFirer<I, S>,
+    E: Executor<EM, I, S, Z>,
+    Z: Evaluator<E, EM, I, S>,
+    I: Input,
+    S: HasExecutions + HasNamedMetadata,
+{
+    // This is only a stat stage, we don't need to restart anything
+    fn should_restart(&mut self, _state: &mut S) -> Result<bool, Error> {
```
LCBH, line 379: Is this how some stages are implemented in libAFL?

puffin/src/fuzzer/utils.rs
```diff
@@ -103,7 +103,7 @@ where
         None
     } else {
         // pick a random, valid index
-        let index = rand.below(length as u64) as usize;
+        let index = rand.below_or_zero(length);
```
LCBH, line 106: Is it really exclusive? (length is never picked)

puffin/src/fuzzer/bit_mutations.rs
```diff
@@ -733,14 +745,24 @@ impl<S, PT> Mutator<Trace<PT>, S> for [<$mutation  DY>]<S>
...
-impl<S> Named for [<$mutation  DY>]<S>
+impl<S, PT> Named for [<$mutation  DY>]<S, PT>
     where
         S: HasRand + HasMaxSize,
+        PT: ProtocolTypes,
 {
-    fn name(&self) -> &str {
-        remove_prefix_and_type(std::any::type_name::<[<$mutation  DY>]<S>>())
+    fn name(&self) -> &Cow<'static, str> {
+        &Cow::Borrowed(stringify!($mutation))
```
LCBH, line 765: Are we sure this will create the same as the short name that we previously produced with `remove_prefix_and_type`?

puffin/src/fuzzer/stats_monitor.rs
```diff
@@ -32,10 +31,12 @@ pub struct StatsMonitor {
...
 impl StatsMonitor {
     pub fn with_tui_output(stats_file: PathBuf) -> Self {
-        let monitor = Box::new(TuiMonitor::new(TuiUI::new(
-            String::from("tlspuffin [press q to exit]"),
-            false,
-        )));
+        let monitor = Box::new(
+            TuiMonitor::builder()
+                .title(String::from("tlspuffin [press q to exit]"))
+                .enhanced_graphics(false)
```
LCBH: I guess we should try with `true ` here and see if it works on our machines. Others can use the standard monitor.

puffin/src/fuzzer/stats_monitor.rs
```diff
@@ -56,12 +57,17 @@ impl StatsMonitor {
...
     }
 
-    fn client(&mut self, id: ClientId) -> Statistics {
-        let client = self.client_stats_mut_for(id);
+    fn client(
+        &mut self,
+        client_stats_manager: &mut ClientStatsManager,
+        id: ClientId,
+    ) -> Statistics {
+        let client = client_stats_manager.client_stats_for(id).unwrap();
```
LCBH: Go for a panic-free solution here (no `unwrap`)

puffin/src/fuzzer/stats_monitor.rs
```diff
@@ -56,12 +57,17 @@ impl StatsMonitor {
...
+        let mut cur_client_clone = client.clone();
```
LCBH: We must avoid this costly `clone`

puffin/src/fuzzer/stats_monitor.rs
```diff
@@ -666,6 +664,7 @@ impl JSONEventHandler {
     {
         let writer = BufWriter::new(
             OpenOptions::new()
+                // The wrong trait is used and breaks the chain if libafl_bolts prelude is imported with *
```
LCBH: What does this comment mean?

puffin/src/fuzzer/stages.rs
```diff
@@ -395,42 +290,55 @@ where
...
     /// Compute the number of iterations used to apply stacked mutations
     fn iterations(&self, state: &mut S, _: &I) -> u64 {
-        state.rand_mut().below(self.max_mutations_per_iteration)
+        state
+            .rand_mut()
+            .below_or_zero(self.max_mutations_per_iteration)
+            .try_into()
+            .unwrap()
```
LCBH: Go for a panic-free solution

LCBH (changes requested):

puffin/src/fuzzer/bit_mutations.rs
nataelbaffou: We get the struct name (ex: BitFlipMutator) as what was done before

puffin/src/fuzzer/mutations.rs
nataelbaffou: Done

puffin/src/fuzzer/stages.rs
nataelbaffou: We can just cast with as, we don't need to worry about truncation from usize to u64

puffin/src/fuzzer/stages.rs
nataelbaffou: done

Cargo.toml
nataelbaffou: I didn't find a workaround except using a vec instead of an array. I created an issue in there github. We'll see where it goes.

puffin/src/fuzzer/utils.rs
nataelbaffou: It is exclusive from length and inclusive with zero. I'm not sure to understand the question

puffin/src/fuzzer/stats_monitor.rs
nataelbaffou: Propagate the error to the caller, it already manages the error

puffin/src/fuzzer/stats_monitor.rs
nataelbaffou: The wrong "append" function is called : libafl::tuples::append instead of std::fs::OpenOptions append when we imported libafl_bolts::prelude with "*"

Changed the code so that we explicitly use the right append function. A bit more wordy but better than adding warning comments.

puffin/src/fuzzer/stats_monitor.rs
nataelbaffou: Done

puffin/src/fuzzer/stats_monitor.rs
nataelbaffou: It works fine, it changes the tui from ascii to unicode

puffin/src/fuzzer/stats_stage.rs
nataelbaffou: Yes AflStatStage uses the same methods/return values

LCBH (approved):

PR #460, opened 2026-02-24
Fix put_use_clear flag

nataelbaffou:
Put options defined in CLI are now in every default PutDescriptor/PutOptions

- Changed put_use_clear attribute to put_options in Fuzzerconfig
- Set default put_options  in CLI to defined put_options (here, only put_use_clear if defined)
- Added support for use_clear in CPUT openssl (used clear by default bu supports free now also, by default)

### DIFF
--- a/puffin/src/cli.rs
+++ b/puffin/src/cli.rs
@@ -20,7 +20,7 @@ use crate::fuzzer::start;
 use crate::graphviz::write_graphviz;
 use crate::log::config_default;
 use crate::protocol::{ProtocolBehavior, ProtocolTypes};
-use crate::put::{PutDescriptor, PutOptions};
+use crate::put::PutDescriptor;
 use crate::put_registry::{PutRegistry, TCP_PUT};
 use crate::trace::{Action, ConfigTrace, ExecutionResult, Source, Spawner, Trace, TraceContext};
 
@@ -171,12 +171,6 @@ where
     asan_info();
     setup_asan_env();
 
-    // Initialize global state
-    let mut options: Vec<(String, String)> = Vec::new();
-    if put_use_clear {
-        options.push(("use_clear".to_string(), put_use_clear.to_string()));
-    }
-
     // Set up config
     let mut config = FuzzerConfig {
         static_seed,
@@ -195,7 +189,9 @@ where
         return ExitCode::FAILURE;
     }
     if put_use_clear {
-        config.put_use_clear = true;
+        config
+            .put_options
+            .add_option("use_clear", put_use_clear.to_string().as_str());
     }
     if with_bit_level {
         config.mutation_config.with_bit_level = true;
@@ -210,6 +206,9 @@ where
     if with_truncation {
         config.mutation_stage_config.with_truncation = true;
     }
+
+    // Set put_options as default for every PutDescriptor created by the put_registry
+    put_registry.set_default_options(config.put_options.clone());
 
     // Setup Logging
     // We need to create the log directory before initializing the logger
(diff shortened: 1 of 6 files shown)


puffin/src/fuzzer/config.rs
```diff
@@ -28,7 +28,9 @@ pub struct FuzzerConfig {
     pub no_launcher: bool,
     pub log_folder: PathBuf,
     pub is_experiment: bool,
-    pub put_use_clear: bool, // use clear instead of free on Agents in between traces of some Input
+    pub put_options: PutOptions, /* Put options to set when creating a put descriptor (ex:use
```
LCBH: Move this comment to `PutOptions` definition

tlspuffin/harness/openssl/src/put.c
```diff
@@ -155,18 +160,63 @@ RESULT openssl_progress(AGENT agent)
     return get_result(agent, ret, true);
 }
 
-RESULT openssl_reset(AGENT agent, uint8_t new_name, uint8_t use_clear)
+static bool recreate_ssl_from_agent_ctx(AGENT agent)
 {
-    agent->name = new_name;
+    SSL_free(agent->ssl); // also frees BIOs (SSL_set_bio transferred ownership)
```
LCBH: Sure about this freeing in and out?

LCBH (changes requested):

puffin/src/cli.rs
```diff
@@ -194,7 +188,9 @@ where
         return ExitCode::FAILURE;
     }
     if put_use_clear {
-        config.put_use_clear = true;
+        config
+            .put_options
+            .add_option("use_clear", put_use_clear.to_string().as_str());
```
LCBH, line 194: This is risky as not really robust to small typos and not really robust/future-proof. I know it is legacy. We shoudld maybe discuss refactoring PutOptions and interface to puffin.

LCBH (changes requested):

puffin/src/cli.rs
```diff
@@ -469,6 +468,7 @@ where
         if let Some(cwd) = cwd {
             options.push(("cwd", cwd));
         }
+        // TODO: Should we use the generic put options here also ? Aka put_use_clear ?
```
LCBH: I don't understand this comment: we already parsed the put_use_clear option right?

puffin/src/cli.rs
```diff
@@ -613,11 +618,11 @@ where
...
 
             FuzzingTarget::Differential(
-                PutDescriptor::new(first_put, PutOptions::empty()),
-                PutDescriptor::new(second_put, PutOptions::empty()),
+                PutDescriptor::new(first_put, put_registry.default_put_options().clone()),
+                PutDescriptor::new(second_put, put_registry.default_put_options().clone()),
             )
         } else {
-            FuzzingTarget::default()
+            FuzzingTarget::Single(None)
```
LCBH, line 646: Why this edit?

LCBH (changes requested):

tlspuffin/harness/openssl/src/put.c
nataelbaffou: Yes, it calls

    BIO_free_all(s->wbio);
    BIO_free_all(s->rbio);

puffin/src/cli.rs
nataelbaffou: We define the type of the fuzzing target, makes it more explicit because that's not a default path if we miss everything but the path taken for each single target experiment.

LCBH (approved):

PR #461, opened 2026-03-02
Tls agent support multiple versions

### LINKED ISSUE #368: tlspuffin : investigate TLS 1.2 encryption/decription function symbols
It seems that we don't have functions symbols to decrypt TLS 1.2 messages.

### LINKED ISSUE #434: [WolfSSL] Support downgrade mechanism of TLS
Currently, our WolfSSL harness supports TLS 1.2 only (`wolfTLSv1_2_client_method`) or TLS 1.3 only (`wolfTLSv1_3_client_method`) agents. This prevents us from fuzzing the version negocation  mechanism of TLS, eg. a TLS 1.2 server negociating with a TLS 1.3 client that also support TLS 1.2., some existing CVE have leveraged this mechanism to cause ciphersuite downgrade  (cf. [CVE-2024-5814](https://nvd.nist.gov/vuln/detail/CVE-2024-5814)).

WolfSSL provides [wolfsslv23_client_method](https://www.wolfssl.com/documentation/manuals/wolfssl/group__Setup.html#function-wolfsslv23_client_method) and [wolfSSLv23_server_method](https://www.wolfssl.com/documentation/manuals/wolfssl/group__Setup.html#function-wolfSSLv23_server_method) along with [wolfSSL_CTX_SetMinVersion](https://www.wolfssl.com/documentation/manuals/wolfssl/group__Setup.html#function-wolfssl_ctx_setminversion) that allow an agent to support a range of TLS version and the TLS version negotiation mechanism.

nataelbaffou:


tlspuffin/tests/vulnerabilities.rs
```diff
@@ -670,3 +668,14 @@ fn tcp_wolfssl_cve_2022_39173() {
     let shutdown = context.find_agent_mut(server).unwrap().shutdown();
     log::info!("{}", shutdown);
 }
+
+#[apply(test_puts,
+     attrs = [should_panic(expected = "Negotiated cipher is not in agent's configured ciphers list")],
+     filter = all(CVE_2024_5814, tls12, tls13, not(feature = "wolfssl430"))
+)] // We don't have the claims for wolf rust put
```
LCBH: specify which claims are missing (the newly introduced one)

puffin/src/fuzzer/harness.rs
```diff
@@ -44,11 +44,18 @@ pub fn harness<PB: ProtocolBehavior + 'static>(
...
-            HARNESS_EXEC_AGENT_SUCCESS.increment();
+    match runner.execute(input, &mut fail_at_step, true) {
+        Ok(ctx) => {
+            HARNESS_EXEC_SUCCESS.increment();
+            if cfg!(feature = "introspection") && ctx.agents_successful() {
+                HARNESS_EXEC_AGENT_SUCCESS.increment();
+            }
+        }
+        Err(Error::SecurityClaim(msg)) => {
+            log::warn!("{}", msg);
```
LCBH, line 61: Can you add a comment before the abort explaining why it is there and why it is critical :)

LCBH (changes requested):
I think we want two PRs: one that allows v1.2 + V1.2 + new seed (so that we can explore the impact of that on the fuzzer) and one that extends mutations with global application + the new CVE test.
Note: I need a bit more work to come up with a clean global mutation PR. But we can already finish the first PR.

LCBH (changes requested):
I think we want two PRs: one that allows v1.2 + V1.2 + new seed (so that we can explore the impact of that on the fuzzer) and one that extends mutations with global application + the new CVE test.
Note: I need a bit more work to come up with a clean global mutation PR. But we can already finish the first PR.

puffin-build/src/harness/put.rs
```diff
@@ -79,6 +79,11 @@ impl Put {
 
     pub fn print_cargo_metadata(&self) {
         println!("cargo:rerun-if-changed={}", self.library.path().display());
+        // Check for harness files
+        println!("cargo:rerun-if-changed={}", self.harness.path.display());
+        // Check for added or removed vendors: slow, adds a few seconds before each run
+        // but can be useful when switching often, building new ones etc...
+        // println!("cargo:rerun-if-changed={}",self.library.path().join("..").display());
```
LCBH: Is it really costly? If yes, maybe add a feature flag for that?

puffin/src/fuzzer/mutations.rs
```diff
@@ -649,44 +650,100 @@ where
...
-                let ctx = TraceContext::new(spawner); // zoo generate symbolic terms
-                self.zoo.insert(TermZoo::generate(
-                    &ctx,
-                    self.signature,
-                    rand,
-                    self.constraints.zoo_gen_how_many,
-                ))
-            } else {
-                self.zoo.get_or_insert_with(|| {
+        let (to_mutate_path, to_mutate, new_term) = {
```
LCBH: FILE: To integrate in new pR

puffin/src/fuzzer/utils.rs
```diff
@@ -6,8 +6,11 @@ use crate::trace::{Action, Step, Trace};
 
 #[derive(Copy, Clone, Debug)]
 pub struct TermConstraints {
+    // For selecting (sub)terms for mutation candidates, for example
```
LCBH: FILE: to include in another PR

puffin-build/vendors/wolfssl/builder.cmake
```diff
@@ -62,7 +66,8 @@ autotools_builder(
     --enable-sp
     --disable-sha3
     --enable-curve25519
-    --enable-secure-renegotiation
+    --enable-renegotiation-indication # Adds a security layer above secure-renegotiation
```
LCBH, line 69: Why do we need these edits?

tlspuffin/src/rust_put/openssl/mod.rs
```diff
@@ -197,7 +197,7 @@ impl RustPut {
 
         // Allow EXPORT in server
         match descriptor.protocol_config.tls_version {
-            TLSVersion::V1_3 => {
```
LCBH, line 200: Are we sure these two methods `set_cipher_list` and `set_cipher_list` have teh same effect? (I know this is not modified here)

tlspuffin/src/rust_put/openssl/util.rs
```diff
@@ -23,13 +23,28 @@ pub fn set_max_protocol_version(
     match tls_version {
         TLSVersion::V1_3 => {
             #[cfg(feature = "openssl111_binding")]
-            ctx_builder.set_max_proto_version(Some(openssl::ssl::SslVersion::TLS1_3))?;
+            {
+                ctx_builder.set_max_proto_version(Some(openssl::ssl::SslVersion::TLS1_3))?;
+                ctx_builder.set_min_proto_version(Some(openssl::ssl::SslVersion::TLS1_3))?;
+            }
```
LCBH, line 30: Why no ` #[cfg(feature = "libressl333")]` and libressl-specific command?

tlspuffin/src/tls/rustls/msgs/message.rs
```diff
@@ -93,6 +93,9 @@ impl MessagePayload {
             ContentType::Handshake => {
                 HandshakeMessagePayload::read_version(&mut r, vers)
                     .map(MessagePayload::Handshake)
+                    // Avoid accepting clear handshake type if there is still data left in the buffer
```
LCBH: This is surprising: do you mean LibreSSL masquerades some TLS1.2 encrypted message as Handshake message?

tlspuffin/src/tls/fn_cert.rs
```diff
@@ -91,7 +95,7 @@ pub fn fn_rsa_sign_client(
     private_key: &Vec<u8>,
     scheme: &SignatureScheme,
 ) -> Result<Vec<u8>, FnError> {
-    _fn_rsa_sign(
```
LCBH, line 94: I think the naming rationale that `fn_*` should be in the signature, which is not the case of this function symbol.

tlspuffin/src/tls/fn_fields.rs
```diff
@@ -253,6 +275,10 @@ pub fn fn_weak_export_cipher_suite() -> Result<CipherSuite, FnError> {
     Ok(CipherSuite::TLS_RSA_EXPORT_WITH_DES40_CBC_SHA)
 }
 
+pub fn fn_excluded_cipher_suite() -> Result<CipherSuite, FnError> {
```
LCBH, line 280: please add a Rust doc, explain why "excluded"

LCBH (changes requested):

tlspuffin/src/rust_put/openssl/mod.rs
nataelbaffou: I do not have the exact effect remembered but they do not manage the ciphers exactly the same way regarding the ciphers versions. I did check that everything worked fine.

tlspuffin/src/rust_put/openssl/util.rs
nataelbaffou: In libressl333 the TLS1_3 enum does not exists, we cannot use it to define 1.3 specific behaviours

tlspuffin/src/tls/rustls/msgs/message.rs
nataelbaffou: No, it just happened that the encrypted message started with for 0x0 bytes and we parsed that as an HelloRequest with no payload (normal) without checking that we still had some payload left behind, it therefore meant that it was an encrypted handshake message

nataelbaffou: Added a specific method "read_version_exact" to check that, we do not have to manage the reader leftover here.

tlspuffin/src/tls/fn_cert.rs
nataelbaffou: Displaced and renamed it and ecdsa_sign to a new file sign.rs, we now only have function symbols in this file.

puffin-build/vendors/wolfssl/builder.cmake
nataelbaffou: To normalize across PUTs, it adds an extension in client/server hello

puffin-build/src/harness/put.rs
nataelbaffou: It really is costly because it watches over all the files recursively in the target directory. I added the feature flag "watch-vendor" to enable it.

tlspuffin/src/tls/mod.rs
```diff
@@ -249,10 +253,13 @@ define_signature!(
...
     fn_get_ticket_nonce [get]
     fn_new_transcript12
-    fn_decode_ecdh_pubkey [opaque]
+    fn_decode_server_ecdh_pubkey [opaque]
+    fn_decode_client_ecdh_pubkey [opaque]
+    fn_sign_rsa_ecdhe_server_key_exchange [opaque]
     fn_encode_ec_pubkey12
     fn_new_pubkey12 [opaque]
     fn_encrypt12 [opaque]
+    fn_decrypt12 [no_gen]
```
LCBH, line 262: Is this `[no_gen]` mandatory?

tlspuffin/Cargo.toml
```diff
@@ -15,6 +15,8 @@ categories.workspace = true
 [features]
 default = ["sancov", "introspection"]
 
+watch-vendor = ["puffin-build/watch-vendor"]
```
LCBH, line 19: Add doc comment

tlspuffin/src/tls/fn_utils.rs
```diff
@@ -700,13 +705,47 @@ pub fn fn_new_transcript12() -> Result<HandshakeHash, FnError> {
...
+}
+
+pub fn fn_decode_client_ecdh_pubkey(data: &Vec<u8>) -> Result<Vec<u8>, FnError> {
+    let mut rd = Reader::init(data.as_slice());
+    let params = ClientECDHParams::read(&mut rd)
+        .ok_or_else(|| FnError::Codec("Failed to parse client ecdh public key".to_string()))?;
     Ok(params.public.0)
 }
 
+pub fn fn_sign_rsa_ecdhe_server_key_exchange(
```
LCBH: Only used in TLS1.2 right?

tlspuffin/src/tls/sign.rs
```diff
@@ -0,0 +1,60 @@
+use std::sync::Arc;
+
+use puffin::algebra::error::FnError;
+use ring::signature::RsaKeyPair;
+
+use crate::tls::rustls::key::PrivateKey;
+use crate::tls::rustls::msgs::enums::SignatureScheme;
+use crate::tls::rustls::sign::{EcdsaSigningKey, RsaSigner, SigningKey};
+
+pub fn rsa_sign(
```
LCBH, line 10: TODO in a new issue: expose this to the attacker (add to signature)

tlspuffin/src/protocol.rs
```diff
@@ -332,9 +333,10 @@ pub struct TLSDescriptorConfig {
     /// Whether we want to try to reuse a previous agent. This is needed for TLS session resumption
     /// as openssl agents rotate ticket keys if they are recreated.
     pub try_reuse: bool,
-    /// List of available TLS ciphers
-    pub cipher_string_tls13: String,
-    pub cipher_string_tls12: String,
+    /// List of available TLS ciphers - public for default() to be used
+    /// Not supposed to be used outside the public methods to manage 1.2/1.3/both
+    pub _cipher_string_tls13: String,
```
LCBH, line 338: why `_` prefix?

LCBH: why pub?

tlspuffin/src/protocol.rs
```diff
@@ -366,19 +368,252 @@ impl TLSDescriptorConfig {
...
+    /// Sets both 1.2 and 1.3 cipher strings. If you need to specify two different cipher strings,
+    /// use `set_cipher_string_12` and `set_cipher_string_13` separately.
+    /// When using tlsversion::both both strings will be concatenated (and deduped).
+    /// Be careful if you set only one of them when using both.
+    pub fn set_cipher_string(&mut self, cipher_string: String) {
+        self._cipher_string_tls13 = cipher_string.clone();
+        self._cipher_string_tls12 = cipher_string;
+    }
+
+    fn concat_cipher_strings(&self) -> String {
```
LCBH, line 389: Is this the way to configure TLS PUts as per the RFC or is this as per the standard OpenSS API? Which PUTs will accept such concats?

LCBH (changes requested):

tlspuffin/src/tls/mod.rs
nataelbaffou: Yes, see test_tils.rs:314 comment  "Those 2 are the function symbols for which we can generate a term but all fail to DY_execute! Indeed, the HandshakeHash that is fed as argument must be computed in a very specific way! We might give known,valid hash-transcript to help?"

tlspuffin/src/tls/fn_utils.rs
nataelbaffou: Yes, renamed with 12 at the end of the function

tlspuffin/src/protocol.rs
nataelbaffou: Not from an RFC, just a way for us to allow to set individually or collectively tls1.2 and 1.3 ciphers and abstract the fact that if we set a client/server at both we need to have all the ciphers in a single string.

tlspuffin/Cargo.toml
nataelbaffou: Added a doc line above and create an issue #479

tlspuffin/src/tls/sign.rs
nataelbaffou: Created issue #478

LCBH (approved):
LGTM

PR #473, opened 2026-03-24
441 mk vendor parse config improvement

### LINKED ISSUE #441: Minor observation
Hi all,
In reading the implementation of `parse_config_arg` of `mk_vendor`
https://github.com/tlspuffin/tlspuffin/blob/0b44eed3b28589f0d9e266e572bfea6e8d894f62/puffin-build/src/tools/mk_vendor.rs#L77-L88
I think there are several minor problems.

First, there is a `s.trim()` to remove the leading and trailing whitespaces of the argument, I suppose that it helps remove whitespaces in "vendor" and "name", for example if the argument is:

```
"       abc:def      "
```
then after parsing, there will be `vendor = "abc"` and `name = "def"`. But if the whitespaces are around the separator ':', for example:

```
"abc     :         def
```
then `vendor = "abc     "` and `name = "         def"`, so the single trimming may not help in this case.

Second, the function uses `Regex` crate to parse the argument, and the regular expression compilation invoked in `Regex::new` is not really negligible, citing from [the doc](https://docs.rs/regex/1.10.6/regex/bytes/struct.Regex.html) 

> Note that regex compilation tends to be a somewhat expensive process, and unlike higher level environments, compilation is not automatically cached for you. One should endeavor to compile a regex once and then reuse it. For example, it’s a bad idea to compile the same regex repeatedly in a loop.

It is absolutely not a critical bottleneck since `parse_config_arg` is called only one time, but it requires also including the crate `Regex` which increases the compilation time (and the binary size) of `mk_vendor`.

IMHO, an alternative implementation which should do the same:

```Rust
fn parse_config_arg(s: &str) -> Result<(String, String), String> {
    match s.split_once(':') {
        Some((vendor, name)) if !vendor.trim().is_empty() && !name.trim().is_empty() && !name.contains(':') => {
            Ok((vendor.trim().to_string(), name.trim().to_string()))
        }
        _ => Err("invalid config format (expected '<vendor>:<name>')".to_string()),
    }
}
```
and is zero allocation, Regex free.

I've benchmarked the two versions using hyperfine, following is the results:

- Current implementation
```
hyperfine "./tools/mk_vendor make "csdssdf:sdfsdfsdfa"" -i --warmup 7 --runs 200
Benchmark 1: ./tools/mk_vendor make csdssdf:sdfsdfsdfa
  Time (mean ± σ):     133.7 ms ±   2.5 ms    [User: 95.7 ms, System: 36.9 ms]
  Range (min … max):   130.0 ms … 143.2 ms    200 runs

  Warning: Ignoring non-zero exit code.
```

- Alternative implementation
```
hyperfine "./tools/mk_vendor make "csdssdf:sdfsdfsdfa"" -i --warmup 7 --runs 200
Benchmark 1: ./tools/mk_vendor make csdssdf:sdfsdfsdfa
  Time (mean ± σ):     131.6 ms ±   1.9 ms    [User: 95.2 ms, System: 36.2 ms]
  Range (min … max):   129.0 ms … 141.0 ms    200 runs

  Warning: Ignoring non-zero exit code.
```
which seems slightly faster.

The notices above are absolutely minor, and the current implementation works perfectly fine. Worse, I may be wrong or missing context, so please feel free to ignore or reject if it doesn’t fit the project’s goals.

Thanks you for any feedback.

nataelbaffou:
Improves mk_vendor as suggested by [tathanhdinh](https://github.com/tathanhdinh) in #441

### DIFF
--- a/puffin-build/src/tools/mk_vendor.rs
+++ b/puffin-build/src/tools/mk_vendor.rs
@@ -1,6 +1,5 @@
 use clap::{Parser, Subcommand};
 use puffin_build::{library, vendor_dir};
-use regex::Regex;
 
 #[derive(Debug, Parser)]
 #[command(about = "helper script to build vendor libraries for tlspuffin", long_about = None)]
@@ -75,14 +74,13 @@ pub fn main() -> std::process::ExitCode {
 }
 
 fn parse_config_arg(s: &str) -> Result<(String, String), String> {
-    let s = s.trim();
-
-    let Some(captures) = Regex::new(r"^(?<vendor>[^:]+):(?<name>[^:]+)$")
-        .unwrap()
-        .captures(s)
-    else {
-        return Err("invalid config format (expected '<vendor>:<name>')".to_string());
-    };
-
-    Ok((captures["vendor"].to_string(), captures["name"].to_string()))
+    match s
+        .split_once(':')
+        .map(|(vendor, name)| (vendor.trim(), name.trim()))
+    {
+        Some((vendor, name)) if !vendor.is_empty() && !name.is_empty() && !name.contains(':') => {
+            Ok((vendor.to_string(), name.to_string()))
+        }
+        _ => Err("invalid config format (expected '<vendor>:<name>')".to_string()),
+    }
 }


nataelbaffou:
Fixes #441

aeyno (approved):
LGTM

LCBH (approved):
LGTM

PR #474, opened 2026-03-24
[build] 440 - library config parsing simplified

### LINKED ISSUE #440: Function requires consuming argument but actually it requires borrowing only
Hello all,

I observe that the `Config` has an `AsRef` implementation https://github.com/tlspuffin/tlspuffin/blob/0b44eed3b28589f0d9e266e572bfea6e8d894f62/puffin-build/src/library/build/config.rs#L72-L76

which is obliged since `config` is consumed by

https://github.com/tlspuffin/tlspuffin/blob/0b44eed3b28589f0d9e266e572bfea6e8d894f62/puffin-build/src/vendor_dir/mod.rs#L149

Actually, the logic of the function `make` requires borrowing `config` only. Indeed:

https://github.com/tlspuffin/tlspuffin/blob/0b44eed3b28589f0d9e266e572bfea6e8d894f62/puffin-build/src/vendor_dir/mod.rs#L150-L151

and

https://github.com/tlspuffin/tlspuffin/blob/0b44eed3b28589f0d9e266e572bfea6e8d894f62/puffin-build/src/vendor_dir/mod.rs#L157-L160

IMHO, it is possible to simplify the function by borrowing `config`, e.g.

```Rust
pub fn make(&self, config: &library::Config, cput: bool) -> Result<PathBuf> {
    if self.contains(config) {
        return Ok(self.path.clone());
    }

    // NOTE ensure intermediate folders exist and there is no artifact from a previous build
    self.remove()?;

    config.build(self.path(), cput).map(|_| {
        std::fs::write(
            self.config_file(),
            toml::to_string(config).unwrap(),
        )
        .ok();
        self.path().to_path_buf()
    })
}
```

then it allows also to remove the `AsRef<Config>` implementation for `Config`.

Many thanks for any response.

nataelbaffou:
Fixes #440

### DIFF
--- a/crates/boringssl-src/src/lib.rs
+++ b/crates/boringssl-src/src/lib.rs
@@ -68,7 +68,7 @@ pub fn build(options: &BoringSSLOptions) -> Artifacts {
 
     let prefix = vendor_dir::from_env()
         .library_dir(&name)
-        .and_then(|dir| dir.make(config, false))
+        .and_then(|dir| dir.make(&config, false))
         .unwrap();
 
     Artifacts {
--- a/crates/libressl-src/src/lib.rs
+++ b/crates/libressl-src/src/lib.rs
@@ -63,7 +63,7 @@ impl Build {
 
         let prefix = vendor_dir::from_env()
             .library_dir(&name)
-            .and_then(|dir| dir.make(config, false))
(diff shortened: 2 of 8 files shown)


aeyno (approved):
LGTM

LCBH (approved):

PR #481, opened 2026-04-16
[TLS:Harness:Wolf] fix C harness: fix memory leaks and incorrect peer cert key type in wolfSSL

odemengeon:


nataelbaffou:
LGTM and tested

nataelbaffou (approved):

PR #487, opened 2026-04-24
[Cli][Puffin-MutationalStage] Fix regression in differential number of objectives

nataelbaffou:
Fixes default value of max-retry=3 for each corpus item in a mutational stage back to unlimited.
Adds a cli parameter to modify this value `--mutational-retries` with default to 0 (unlimited)


aeyno (approved):
Tested, fixes regression.

PR #488, opened 2026-05-07
Objectives created with feedback instead of crashing

nataelbaffou:
Switch Security claims and differential differences from crashing to sending a feedback to create an objective.

### DIFF
--- a/puffin/src/fuzzer/feedback.rs
+++ b/puffin/src/fuzzer/feedback.rs
@@ -14,10 +14,62 @@ use serde::{Deserialize, Serialize};
 use crate::protocol::ProtocolTypes;
 use crate::trace::Trace;
 
-// A global (or thread-local) mutable variable that your harness will update.
-// Now it holds an Option<usize>.
 thread_local! {
     pub static FAIL_AT_STEP: Cell<Option<usize>> = const { Cell::new(None) };
+    pub static OBJECTIVE_TRIGGERED: Cell<bool> = const { Cell::new(false) };
+}
+
+/// Feedback that triggers when the harness sets [`OBJECTIVE_TRIGGERED`] to `true`.
+///
+/// Used to record security-claim violations and differential differences as objectives
+/// without crashing the process (avoids the restart + serialization overhead of abort()).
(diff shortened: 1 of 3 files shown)


aeyno (approved):
LGTM

puffin/src/fuzzer/harness.rs
LCBH, line 77: Do we want to return `ExitKind::Ok` in the case of ` Err(Error::SecurityClaim(msg))`?

LCBH, line 77: Same as above

LCBH (changes requested):

puffin/src/fuzzer/feedback.rs
```diff
@@ -14,10 +14,62 @@ use serde::{Deserialize, Serialize};
...
+{
+    fn is_interesting(
+        &mut self,
+        _: &mut S,
+        _: &mut EM,
+        _: &I,
+        _: &OT,
+        _: &ExitKind,
+    ) -> Result<bool, Error> {
+        Ok(OBJECTIVE_TRIGGERED.get())
```
LCBH, line 60: What if the harness has REALLY crashed? Do we still consider the trace as an obejctive?

LCBH (changes requested):

puffin/src/fuzzer/feedback.rs
nataelbaffou: Yes

puffin/src/fuzzer/harness.rs
nataelbaffou: Yes we have to return OK to not trigger any other feedback or restarts from happening

puffin/src/fuzzer/feedback.rs
LCBH: Ah yes, thanks to this in libafl_setup:
```
                ObjectiveFeedback::new(),
                CrashFeedback::new(),
                TimeoutFeedback::new()
```

Got it!

LCBH (approved):

PR #491, opened 2026-05-19
Deduplicate differential objectives

aeyno:
This PR updates the `ObjectiveFeedback` to deduplicate some of the objectives of differential fuzzing.

Every objective gets assigned a  hash containing the differences, put errors and corpus trace id: traces with the same differences and put errors originating from the mutation of the same corpus trace as a previously found objective should be deduplicated.

This is a very conservative approach and should not reduce the diversity of objectives. After some first experiments this divides the number of objectives by at least 3 and this proportion increases with the saturation of the corpus.

The deduplication mechanism is currently applied to differential fuzzing but could be extended to standard DY fuzzing provided that the fuzzer create a hash of an objective.

The PR also adds a `duplicate` stat (global and per client).

### DIFF
--- a/puffin/src/execution.rs
+++ b/puffin/src/execution.rs
@@ -284,21 +284,27 @@ impl<PB: ProtocolBehavior> TraceRunner for &DifferentialRunner<PB> {
 
         // If we have status diff or security violation we return the diffs without checking for
         // others diffs
-        if !diff.is_empty() {
-            return Err(Error::Difference(diff));
-        }
+        if diff.is_empty() {
+            // Compare the trace context
+            diff.extend(first_ctx.compare(&second_ctx, &trace.as_ref().descriptors));
 
-        // Compare the trace context
-        diff.extend(first_ctx.compare(&second_ctx, &trace.as_ref().descriptors));
-
-        // Apply filter to remove false positives
-        diff = diff
+            // Apply filter to remove false positives
+            diff = diff
             .into_iter()
             .filter(<<PB as ProtocolBehavior>::ProtocolTypes as ProtocolTypes>::differential_fuzzing_filter_diff)
             .collect();
+        }
 
         if !diff.is_empty() {
-            return Err(Error::Difference(diff));
+            return Err(Error::Difference {
+                differences: diff,
+                put1_status: first_trace_status
+                    .err()
+                    .map_or("Success".into(), |x| x.to_string()),
+                put2_status: second_trace_status
+                    .err()
+                    .map_or("Success".into(), |x| x.to_string()),
+            });
         }
 
         Ok((first_ctx, second_ctx))
--- a/puffin/src/fuzzer/feedback.rs
+++ b/puffin/src/fuzzer/feedback.rs
@@ -2,21 +2,24 @@ use std::borrow::Cow;
 use std::cell::Cell;
 use std::default::Default;
 
-use libafl::corpus::Testcase;
+use libafl::corpus::{Corpus, CorpusId, Testcase};
 use libafl::events::EventFirer;
 use libafl::executors::ExitKind;
 use libafl::feedbacks::{Feedback, StateInitializer};
 use libafl::observers::ObserversTuple;
 use libafl::prelude::{HasMaxSize, HasRand};
+use libafl::state::HasCorpus;
 use libafl_bolts::{Error, Named};
 use serde::{Deserialize, Serialize};
 
+use crate::fuzzer::stats_stage::DUPLICATES;
 use crate::protocol::ProtocolTypes;
 use crate::trace::Trace;
 
 thread_local! {
     pub static FAIL_AT_STEP: Cell<Option<usize>> = const { Cell::new(None) };
     pub static OBJECTIVE_TRIGGERED: Cell<bool> = const { Cell::new(false) };
+    pub static OBJECTIVE_HASH: Cell<Option<u64>> = const { Cell::new(None)};
 }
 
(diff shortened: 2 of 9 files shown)


puffin/src/execution.rs
```diff
@@ -284,21 +284,27 @@ impl<PB: ProtocolBehavior> TraceRunner for &DifferentialRunner<PB> {
...
         if !diff.is_empty() {
-            return Err(Error::Difference(diff));
+            return Err(Error::Difference {
+                differences: diff,
+                put1_status: first_trace_status
+                    .err()
+                    .map_or("Succes".into(), |x| x.to_string()),
+                put2_status: second_trace_status
+                    .err()
+                    .map_or("Succes".into(), |x| x.to_string()),
```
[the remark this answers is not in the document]
aeyno: Done

puffin/src/fuzzer/harness.rs
```diff
@@ -170,6 +176,13 @@ pub fn differential_harness<PB: ProtocolBehavior + 'static>(
                         .collect::<Vec<String>>()
                         .join("\n")
                 );
+                let mut h = DefaultHasher::new();
+                diffs.hash(&mut h);
+                s1.hash(&mut h);
+                s2.hash(&mut h);
+
+                OBJECTIVE_HASH.set(Some(h.finish()));
```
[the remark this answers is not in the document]
aeyno: Done

puffin/src/fuzzer/feedback.rs
```diff
@@ -2,33 +2,40 @@ use std::borrow::Cow;
...
+    pub static OBJECTIVE_HASH: Cell<Option<u64>> = const { Cell::new(None)};
 }
 
 /// Feedback that triggers when the harness sets [`OBJECTIVE_TRIGGERED`] to `true`.
 ///
 /// Used to record security-claim violations and differential differences as objectives
 /// without crashing the process (avoids the restart + serialization overhead of abort()).
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub struct ObjectiveFeedback;
+pub struct ObjectiveFeedback {
```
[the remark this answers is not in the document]
LCBH: Could be problematic indeed, ideally we would restore seen_objectives as before?!

puffin/src/fuzzer/feedback.rs
```diff
@@ -2,33 +2,41 @@ use std::borrow::Cow;
...
+    #[serde(skip)]
```
LCBH: remove skip

puffin/src/fuzzer/feedback.rs
```diff
@@ -48,27 +56,53 @@ impl<EM, I, OT, S> Feedback<EM, I, OT, S> for ObjectiveFeedback
...
-        _: &mut S,
+        state: &mut S,
         _: &mut EM,
         _: &I,
         _: &OT,
         _: &ExitKind,
     ) -> Result<bool, Error> {
-        Ok(OBJECTIVE_TRIGGERED.get())
+        if OBJECTIVE_TRIGGERED.get() {
+            match OBJECTIVE_HASH.get() {
```
LCBH, line 73: Ideally, we would instead store the hash in the input metadata. This seems more robust than using the Cell, which is subject to race conditions. Have you considered doing so?

puffin/src/fuzzer/stats_monitor.rs
```diff
@@ -197,9 +207,10 @@ impl Display for Statistics {
             Self::Client(client_stats) => {
                 write!(
                     f,
-                    "(CLIENT) corpus: {}, obj: {}, execs: {}, exec/sec: {}",
+                    "(CLIENT) corpus: {}, obj: {}, duplicates: {}, execs: {}, exec/sec: {}",
```
LCBH: This will "polute" stdout for all fuzzing campaigns, including non differential ones. Consider doing this only for diff campaigns.

LCBH (changes requested):

puffin/src/fuzzer/feedback.rs
aeyno: Done

puffin/src/fuzzer/feedback.rs
aeyno: Discussed offline. We have no other real possibility to pass information about the trace to the feedback (where we also have access to the corpus trace used)

puffin/src/fuzzer/stats_monitor.rs
aeyno: Done

aeyno: Done

aeyno:
@LCBH can we merge this PR ?

puffin/src/fuzzer/feedback.rs
```diff
@@ -2,33 +2,40 @@ use std::borrow::Cow;
...
-pub struct ObjectiveFeedback;
+pub struct ObjectiveFeedback {
+    seen_objectives: std::collections::HashSet<(String, u64)>,
+}
 
 impl ObjectiveFeedback {
     pub fn new() -> Self {
-        Self
+        Self {
+            seen_objectives: std::collections::HashSet::new(),
```
nataelbaffou, line 37: Do we have a fix to reload the serialized seen_objective after a crash ? (instead of using #[serde(skip)]) ?

aeyno: I think they are reloaded after a crash, `#[serde(skip)])` has been removed

LCBH (approved):
LGTM

PR #494, opened 2026-05-26
[tlspuffin][OpenSSL] fix two memory leaks in OpenSSL PUT

aeyno:
This fixes two memory leaks in the OpenSSL C harness.


nataelbaffou (approved):
LGTM

PR #495, opened 2026-05-27
[tlspuffin][LibreSSL] fix two memory leaks in LibreSSL PUT

aeyno:
This fixes two memory leaks in the LibreSSL C harness.


nataelbaffou (approved):
LGTM

PR #524, opened 2026-08-27
[puffin:zoo] Change term zoo generation

aeyno:
This PR changes the way we generate a term zoo :

Previously term zoo chose a root symbol then tried to recursively
fill its arguments until every argument was completed of max depth was
reached.

With a finer grained signature the maximum allowed height of 8 was to
small but increasing the height means increasing the maximum size of
terms exponentially. This lead to more flaky tests and generated terms
over 1000 symbols long which was not manageable.

So the idea was to introduce a MAX_SIZE budget to limit the number of
symbols in the term. But with the previous brute-force approach the
budget was quickly exhausted with some symbols.

So we introduce a new mechanism to guide the term generation to maximize
the chance of creating a valid symbol:

1) We precompute a minimum height and budget for each symbol: constants
are 1/1 and each function with parameters has height 1+max((min of
all symbols producing type T) for type T of each arg) and cost is 1 +
sum((min of all symbols producing type T) for type T of each arg).
2) When building a term we first reserve the minimal budget for each of
its arguments that is reserved to ensure that the global budget will not
be entirely consumed before reaching its last argument.
3) For each of the arguments select uniformely the one producing the
correct type that fits in the budget
4) Repeat recursively for each argument

this algorithm makes the test less flaky and drastically reduce the
chances of failing to construct a term. Though it doesn't prevent a term
from failing evaluation

The PR also introduce the `mapper` cli subcommand:
- `mapper symbols` lists all symbols in the signature
- `mapper signature` lists all symbols along with their signature
- `mapper buildable` lists the cost (heigt and number of symbols) to build each symbol in the signature and if the symbol could be built


puffin/src/cli.rs
```diff
@@ -823,3 +836,119 @@ fn check_if_puts_exist<'a, 'b, PB: ProtocolBehavior>(
...
+    };
+
+    let sig = PT::signature();
+    let mut functions: Vec<_> = sig.functions.iter().collect();
+    // Always sort by short name for readability
+    functions.sort_by(|(a, _), (b, _)| {
+        display(a.name)
+            .cmp(&display(b.name))
+            .then_with(|| a.name.cmp(b.name))
+    });
```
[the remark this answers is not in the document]
aeyno: It compiles and works fine.

puffin/src/algebra/signature.rs
```diff
@@ -24,6 +25,60 @@ pub struct Signature<PT: ProtocolTypes> {
...
+
+        let nth = rand.below_or_zero(fitting);
+        deep_enough
+            .iter()
+            .enumerate()
+            .filter(|(_, cost)| cost.size <= size)
+            .map(|(index, _)| index)
+            .nth(nth)
+            .map(|index| &self.functions[index])
+    }
```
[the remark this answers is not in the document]
aeyno: done

puffin/src/fuzzer/utils.rs
```diff
@@ -132,6 +136,7 @@ impl TermConstraints {
             not_readable: false,
             must_be_det: false,
             zoo_gen_how_many: usize::MAX,
+            zoo_max_depth: DEFAULT_MAX_DEPTH,
```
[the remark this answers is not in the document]
LCBH: ?

puffin/src/fuzzer/term_zoo.rs
```diff
@@ -10,8 +10,48 @@ use crate::fuzzer::utils::Choosable;
...
+/// *encoding* depth of the signature: with constructors generated from the protocol types, a
+/// message is built through a chain of wrappers (for TLS: `fn_message` -> `fn_messagepayload_*`
+/// -> `fn_handshakemessagepayload` -> `fn_handshakepayload_*` -> the payload struct) before the
+/// interesting fields are even reached, so a value that used to sit at depth 2 now sits at depth
+/// 6. A budget sized for a flat signature silently makes whole families of symbols
+/// ungeneratable — the generation attempts do not fail loudly, they just always return `None`.
+///
+/// The signature wants roughly 14 here: a `ClientHello` carrying even one extension needs depth 9,
+/// so at 8 the zoo can only build handshake messages with empty extension lists.
+pub const DEFAULT_MAX_DEPTH: u16 = 14;
```
LCBH: I think such values are better defined at protocol-level, with reasonable default in puffin. No? Also a rationale on how to choose them would be useful!

LCBH: I know it was global before, but as we keep pushing these limits due to protocol Mapper, maybe it makes more sense to make a protocol-specific decision here

puffin/src/fuzzer/term_zoo.rs
```diff
@@ -10,8 +10,48 @@ use crate::fuzzer::utils::Choosable;
...
+/// expensive to exhaust than it used to be when a try could fail for free, which is why the
+/// symbols flagged `no_gen` — whose evaluation never succeeds, so they exhaust it in full — are
+/// now skipped whichever way [`TermZoo::generate_many`] is called.
+///
+/// The budget is generous because a few symbols need it: the *first* term is cheap for every
+/// symbol (`fn_encrypt12`, the slowest, takes under a second), but filling a whole quota for the
+/// likes of `fn_derive_psk`, `fn_fill_binder` or `fn_get_client_key_share` costs seconds. The
+/// fuzzer pays that once every `fresh_zoo_after` mutations (100_000 by default), so it is
+/// amortised there; the tests, which build a zoo per symbol, feel it more.
+const MAX_TRIES: usize = 140_000;
```
LCBH: have you measures whether this could be reduced without impacting zoo expressiveness (see zoo uni tests)?

puffin/src/fuzzer/term_zoo.rs
```diff
@@ -10,8 +10,48 @@ use crate::fuzzer::utils::Choosable;
...
+/// How large, in nodes, a generated term may get.
+///
+/// A depth budget does not bound the size: every argument may recurse as deep as the budget
+/// allows, so the size grows exponentially with the depth (unbounded, a depth of 14 produced terms
+/// of up to 1198 nodes against 43 at a depth of 8). This is the bound that actually keeps terms
+/// small, which lets the depth budget be about reachability alone. Well above the sizes the zoo
+/// produces in practice (a 99th percentile of ~20 nodes), and below the
+/// [`crate::fuzzer::utils::TermConstraints::max_term_size`] beyond which the fuzzer would not
+/// select the term for mutation anyway.
+const MAX_SIZE: usize = 64;
```
LCBH: As before: is this Mapper-dependent? Has the value choice been evidence-based?

puffin/src/fuzzer/utils.rs
```diff
@@ -55,6 +58,7 @@ impl Default for TermConstraints {
                                    * `test_term_payloads_eval`, making sure we successfully
                                    * generate, MakeMessage,
                                    * and evaluate after 10 expansions of TermZoo. Was 1 initially */
+            zoo_max_depth: DEFAULT_MAX_DEPTH,
```
LCBH: we definitely want to reconsider these values now that the ZOO is way less costly: especially zoo_gen_how_many, can we make 100 for example?

LCBH (changes requested):

LCBH:
If not tested yet: see if fn_derive_psk still needs to be in unstable_functions

LCBH (changes requested):
looks good but: fix minor issues above + more evaluation on the size parameters + fn_derive_psk

puffin/src/fuzzer/term_zoo.rs
aeyno: moved to ProtocolBehavior

puffin/src/fuzzer/term_zoo.rs
aeyno: If I remember correctly most terms produced are around 28 nodes (but some cases like lists where exploding and reaching 1000+ nodes) so 64 is a value that allow to build all terms with some margin

puffin/src/fuzzer/term_zoo.rs
aeyno: Some symbols still fail to get 10 evaluable terms so reducing this value will potentially prevent some symbols from being in the term zoo

puffin/src/fuzzer/utils.rs
aeyno: zoo is not less costly : before  you would have failed because 1) you could not create a valid term without exceeding max depth or 2) the produced term could not evaluate. 1) is solved, so now you could only fail if a term could not be evaluated, and evaluation is way more costly than generation. For symbols that are hard to make correct, before we used a subtential part of the 140k tries failing generation, now we often end up with 140k failed evaluations which costs more computing time than before.

As we discussed IRL, there are probably some optimisations in terms of term zoo generation to be done, but just increasing how_many to 100 will result in a lot a time spend generating the term zoo.

aeyno:
`fn_derive_psk` needs to remain in unstable functions, it still does not evaluate all the time

PR #530, opened 2026-09-08
[Long-range] fuzzer: grammar-aware mutation stacking, scope weights, step-locked stacking

LCBH:
## Summary

Grammar-aware mutation stacking for the DY (term-tree) mutation stage, plus two
executability-oriented refinements. Three changes, two commits, on top of `pr/mutations/scoped-mutations`:

1. **Geometric mutation stacking** — replace AFL's log-uniform stack size (`1<<(1+rand(0..7))`,
   mean ~36, tuned for flat byte inputs) with a truncated-geometric one concentrated on 1–2 semantic
   edits (mean ~2, light tail to 16). Heavy stacking destroys deep term structure and stores nearly
   every edit as a coverage hitchhiker.
2. **Scope weights inverse to breadth** — `ScopeWeights` default `global:1, step:2, individual:3`,
   favouring surgical edits over trace-wide rewrites.
3. **Step-locked stacking (default on, `--no-step-lock`)** — confine every anchor selected by the
   mutations stacked within one stage to a single randomly-chosen step, so a stage's coverage gain
   and any executability break are attributable to the same step.

## Why (measured)

Replaying finished corpora showed **~51% of saved traces are broken on replay**, breaks clustered a
few steps after the handshake: with heavy stacking a stage anchors mutations on *different* steps —
an early edit gains coverage while a later edit breaks a subsequent step, yet the broken-tailed trace
is still saved because coverage improved. `stats.json` corroborated it: exec-completion ~24–30% under
AFL-style stacking vs ~79% under geometric stacking.

## Mechanism (step-lock)

- Process-local step lock read in `reservoir_sample` — the single chokepoint every `choose*` helper
  routes through. `StepLockedStackMutator` picks one step per stage and sets the lock around the
  inner stacking loop.
- Global/Step **fan-out** (`find_all_*`) deliberately ignores the lock, so a Global mutation still
  spreads its *effect* trace-wide — only its *anchor* is confined to the locked step.
- Safe as process-local state: LibAFL runs a single mutation thread per fuzzer process (one process
  per core). Alternative designs (state-metadata plumbing, explicit args) are discussed in
  `STEP_LOCK_EVALUATION.md`.

Scope (v1): the lock applies to term-anchor mutations (Swap, RemoveAndLift, ReplaceMatch,
ReplaceReuse, Generate). Whole-step structural mutations (Skip, Repeat) are left unlocked.

## Config / CLI

- `MutationStageConfig.step_locked_stacking` (default `true`); disable with `--no-step-lock`.
- `experiment.rs` adds a `_nolock` run-dir suffix when disabled.

## Tests

- `fuzzer::utils::tests::test_step_lock_confines_anchors_to_locked_step` — anchors never escape the
  locked step, and spread across steps when unlocked.
- Full `fuzzer::` suite passes.

## Evaluation

`STEP_LOCK_EVALUATION.md` (included) defines the methodology: LOCK vs BASE (`--no-step-lock`) arms,
legit-seed reachability of the bad-switch OOB as the headline metric, corpus broken-rate (H1),
throughput/coverage no-harm (H3), N replicates with median+range. An evaluation is currently running;
merge is gated on its outcome.

### DIFF
--- a/STEP_LOCK_EVALUATION.md
+++ b/STEP_LOCK_EVALUATION.md
@@ -1,0 +1,101 @@
+# Step-locked mutation stacking — design & evaluation methodology
+
+## 1. What this feature does
+
+A mutational **stage** may stack several mutations (see `GeometricStackMutator`). With step-locking
+**enabled (default)**, all mutations stacked within one stage confine their **anchor selection** to a
+**single step chosen at random once per stage**. Scope still governs how far the *effect* spreads:
+
+- **Individual** → the one locked-step occurrence,
+- **Step** → all occurrences within the locked step,
+- **Global** → all occurrences across the whole trace (its *effect* reaches other steps, but its
+  *anchor* was chosen inside the locked step).
+
+Disable with the CLI flag **`--no-step-lock`**.
+
+### Rationale (data-backed)
+Replaying finished corpora showed **~51% of saved traces are broken on replay**, with breaks
+clustered a few steps *after* the handshake. Cause: heavy stacking lets one stage anchor mutations
+on **different** steps — an early edit gains coverage while a later edit breaks executability, yet
+the whole broken-tailed trace is still saved because coverage improved (the **coverage-hitchhiker**
+effect). Locking every anchor of a stage to one step makes the coverage gain and any executability
+break attributable to the **same** step.
+
+### Implementation map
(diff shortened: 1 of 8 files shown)


LCBH:
Subsumed by https://github.com/tlspuffin/tlspuffin/pull/534

PR #533, opened 2026-09-11
[puffin:mutations] Scoped global replacement with a single centralized cap check

LCBH:
## Summary
Generalize global replacement into a configurable `MutationScope` (Global / Step / Individual),
applied uniformly through a single `apply_to_targets` path, with **one centralized post-mutation
cap check** (`replacement_within_caps` / `payloads_within_caps`) covering every mutator — replacing
the per-mutator inline size guards.

This combines the best of #531 and the earlier scoped-mutations work into one clean commit:
- **Selection cap** `max_term_size = 300` (as in #531).
- **Per-step result cap = 500** — matches #531's post-mutation rule (`resulting step recipe ≤ 500`),
  generalized to the multi-target scope case (`cur + n·delta ≤ 500` per affected step).
- **Whole-trace cap** `max_result_trace_size = 10000`, kept as an inert safety net — it should not
  be hit in normal operation, so it does not affect evaluation (keeps this PR isolated).
- **Scope weights** inverse to breadth (`global:1, step:2, individual:3`) so surgical edits dominate.
- **Per-term payload budget** enforced centrally in `apply_to_targets`.

Global mutations still apply *uniformly* (all structurally-equal occurrences); only the cap check is
now centralized and checked once, efficiently (only affected steps are re-measured).

## Base
Stacked on #531 (`pr/mutations/fix-max-term-size`).

## Tests
`cargo test -p puffin --lib fuzzer::` passes; `tlspuffin` builds.

### DIFF
--- a/puffin/src/fuzzer/mutations.rs
+++ b/puffin/src/fuzzer/mutations.rs
@@ -4,9 +4,9 @@ use libafl::prelude::*;
 use libafl_bolts::prelude::*;
 
 use super::utils::{
-    choose, choose_iter, choose_term, choose_term_filtered_mut, choose_term_mut,
-    choose_term_path_filtered, find_all_term_filtered, find_term_mut, reservoir_sample, Choosable,
-    TermConstraints,
+    choose, choose_iter, choose_term, choose_term_filtered_mut, choose_term_path_filtered,
+    find_all_sub_term_filtered, find_all_term_filtered, find_term, find_term_mut, reservoir_sample,
+    Choosable, TermConstraints, TracePath,
 };
 use crate::algebra::atoms::Function;
 use crate::algebra::signature::Signature;
@@ -14,12 +14,12 @@ use crate::algebra::{DYTerm, Subterms, Term, TermType};
 use crate::fuzzer::term_zoo::TermZoo;
 use crate::protocol::{ProtocolBehavior, ProtocolTypes};
 use crate::put_registry::PutRegistry;
-use crate::trace::{Spawner, Trace, TraceContext};
+use crate::trace::{Action, Spawner, Trace, TraceContext};
 
 #[derive(Clone, Copy, Debug)]
 pub struct MutationConfig {
     pub fresh_zoo_after: u64,
-    pub max_trace_length: usize,
+    pub max_result_trace_length: usize,
     pub min_trace_length: usize,
     /// Below this term size we no longer mutate. Note that it is possible to reach
     /// smaller terms by having a mutation which removes all symbols in a single mutation.
@@ -29,6 +29,9 @@ pub struct MutationConfig {
     pub with_dy: bool,
     /// Focus on one payload at a time for a whole StdMutationalStage
     pub with_focus: bool,
+    /// Relative weights for the scope at which a *replacement* mutation is applied, see
+    /// [`ScopeWeights`] and [`MutationScope`].
+    pub scope_weights: ScopeWeights,
 }
 
 impl Default for MutationConfig {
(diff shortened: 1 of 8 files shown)


puffin/src/fuzzer/mutations.rs
```diff
@@ -42,10 +45,365 @@ impl Default for MutationConfig {
...
+            Some(Action::Input(input)) => input.recipe.size(),
+            _ => 0,
+        };
+        if cur + count * delta > constraints.max_result_term_size {
+            return false;
+        }
+    }
+
+    // Whole-trace check: old size + total growth. No per-step scan of the *unaffected* steps.
+    let new_trace_size = trace.size() + targets.len() * delta;
```
LCBH: This trace.size could be costly as already computed by the caller no?

puffin/src/fuzzer/mutations.rs
```diff
@@ -42,10 +45,365 @@ impl Default for MutationConfig {
...
+    weights: ScopeWeights,
+    rand: &mut R,
+) -> Vec<TracePath> {
+    scoped_targets_with_cutoff(
+        trace,
+        to_mutate_path,
+        representative,
+        weights,
+        rand,
+        // Unbounded search: the result-size caps (applied in `apply_to_targets`) are the guard
```
LCBH: so why do we need this argument here?

puffin/src/fuzzer/mutations.rs
```diff
@@ -42,10 +45,365 @@ impl Default for MutationConfig {
...
+
+/// [`scoped_targets`] with an explicit `max_term_size_explore` cutoff, so that tests can exercise
+/// the case where the search for the other occurrences gives up.
+fn scoped_targets_with_cutoff<PT: ProtocolTypes, R: Rand>(
+    trace: &Trace<PT>,
+    to_mutate_path: &TracePath,
+    representative: &Term<PT>,
+    weights: ScopeWeights,
+    rand: &mut R,
+    max_term_size_explore: usize,
```
LCBH: why not taking MutationCOnfig in argument, I fuess ScopeWeights is there too?

puffin/src/fuzzer/mutations.rs
```diff
@@ -42,10 +45,365 @@ impl Default for MutationConfig {
...
+) -> Vec<TracePath> {
+    let scope = MutationScope::choose(weights, rand);
+    let (chosen_step, _) = to_mutate_path;
+
+    // the type shape is only compared first to speed up the comparison
+    let filter = |term: &Term<PT>| {
+        term.get_type_shape() == representative.get_type_shape() && term == representative
+    };
+    let constraints = TermConstraints {
+        max_term_size_explore,
```
LCBH: Now that we are the assurance that terms size is capped for real, we good remove this cap. No mor max_term_size_explore. And when searcghing for global and steo targets we can do it with no_cinstraint!

puffin/src/fuzzer/utils.rs
```diff
@@ -39,10 +89,13 @@ impl Default for TermConstraints {
     fn default() -> Self {
         Self {
             min_term_size: 0,
-            max_term_size: 300, // we do not select larger terms for mutations
-            /* was 9000 but we were rewriting this to 300 anyway when
-             * instantiating the fuzzer */
+            // Selection cap. Rule: >= 2 * max_seed_term (measured max seed term = 324, TLS).
```
LCBH: keep the value as decided but adapt the comment: we want to be conservative in this PR and we will change these values in a later PR and measure the impact thereof in isolation

puffin/src/fuzzer/utils.rs
```diff
@@ -4,13 +4,63 @@ use crate::algebra::{DYTerm, Term, TermType};
...
+/// | max single-step recipe (nodes) | 324  | 137    | `max_term_size >= 2 * max_seed_term`               |
+/// | max total trace (nodes)        | 427  | 712    | `max_result_trace_size >= ~7 * max_seed_trace`     |
+/// | max #steps                     | 13   | 10     | `max_trace_length >= max_seed_steps + 2`           |
+///
+/// - `max_term_size` MUST be at least `2 * (largest single-term size across all seeds)` so seeds
+///   (and moderately grown variants) stay selectable. Largest today = 324 (TLS) ⇒ rule wants ≥648;
+///   the current value is 650, satisfying the 2× rule. Re-check when seeds/signatures change.
+/// - `max_result_trace_size` MUST comfortably exceed the largest seed trace and a full
+///   `max_trace_length`-step attack; keep it well below where per-execution cost hurts.
+/// - `max_result_term_size` should exceed the largest seed step (324) with headroom; 1000 ≈ 3×.
```
LCBH: keep the value as decided but adapt the comment: we want to be conservative in this PR and we will change these values in a later PR and measure the impact thereof in isolation

tlspuffin/tests/mutations.rs
```diff
@@ -497,7 +498,7 @@ fn search_for_seed_cve_2021_3449(state: &mut TLSState) -> Option<Trace<TLSProtoc
     let mut success = false;
 
     // Check if we can append another encrypted message
-    let mut mutator = RepeatMutator::new(15, true);
+    let mut mutator = RepeatMutator::new(15, usize::MAX, true);
```
LCBH: avoid having usize::Max here!

puffin/src/fuzzer/mutations.rs
```diff
@@ -863,7 +1154,7 @@ mod tests {
     fn test_repeat_mutator() {
         let mut state = create_state();
 
-        let mut mutator = RepeatMutator::new(15, true);
```
LCBH, line 866: Do not put useize::MAX here, and use a MjtationConfig value instead of built-im 15 here, I guess it is a new max_result_trace_length to be added to the struct (and remove max_result_trace_size globally).

puffin/src/fuzzer/utils.rs
```diff
@@ -4,13 +4,65 @@ use crate::algebra::{DYTerm, Term, TermType};
...
 pub struct TermConstraints {
-    // For selecting (sub)terms for mutation candidates, for example
+    /// Minimum size of a sub-term selected as a mutation candidate.
     pub min_term_size: usize,
+    /// Maximum size of a sub-term selected as a mutation candidate. Rule: `>= 2 * max_seed_term`
+    /// (largest single-term size across all seeds). Measured max seed term = 324 (TLS).
     pub max_term_size: usize,
-    // For continuing exploring (sub)terms, if larger: we don't even bother traversing the term
+    /// While sampling, we do not bother traversing a term larger than this.
     pub max_term_size_explore: usize,
```
LCBH: max_term_size_explore should be removed globally right?!

puffin/src/fuzzer/mutations.rs
```diff
@@ -575,6 +920,9 @@ where
     S: HasRand,
 {
     max_trace_length: usize,
+    /// Result-size cap: a duplication that would push the whole trace over this node count is
+    /// rejected (see `TermConstraints::max_result_trace_size`).
+    max_trace_size: usize,
```
LCBH: remove this we dont need it any more right? 

But we need max_result_trace_length in MutationConfig

aeyno (approved):
LGTM

PR #534, opened 2026-09-11
[puffin:mutations] Grammar-based number of mutations per stage + optional step-lock

LCBH:
## Summary
Two mutation-scheduling changes on top of the scoped-replacement PR (#533): a new **default** for how many mutations each stage applies (change 1, **on by default**), and an **opt-in** step-lock option (change 2, **off by default**).

**1. Grammar-based number of mutations per stage (new default).** Replace AFL's log-uniform stack size (`1<<(1+rand(0..7))`, mean ~36, tuned for flat byte inputs) with a truncated-geometric distribution concentrated on 1–2 semantic edits (`GeometricStackMutator`; mean ~2, light tail to 16). Heavy stacking destroys deep DY term structure and saves most edits as coverage hitchhikers. `FixedStackMutator` is also provided to force an exact stack size for experiments.

**2. `--step-lock` (only-one-step) option, disabled by default.** Confine every anchor selected by the mutations stacked within one stage to a single randomly-chosen step, so a stage's coverage gain and any executability break are attributable to the same step (`StepLockedStackMutator` + a process-local step lock read in `reservoir_sample`). Global/Step fan-out still spreads a mutation's effect trace-wide; only the anchor is confined. Ships **off by default** (`--step-lock` to enable); `STEP_LOCK_EVALUATION.md` documents the design and evaluation methodology.

## Base
Stacked on #533 (`pr/mutations/scope-mutations`).

## Tests
`cargo test -p puffin --lib fuzzer::` passes (incl. `test_step_lock_confines_anchors_to_locked_step`); `tlspuffin` builds.

### Note on step-lock vs. a truncation post-filter (future work)
We leave step-lock **off by default for now**: it needs better evaluation, and we want to investigate whether we prefer it and/or a **truncation post-filter** — after mutation *and execution*, before adding to the corpus, restore all steps **after the last failing step** to their pre-mutation terms (those mutations could not have contributed to the positive feedback signal). We also still need to decide what to do with **global** mutations in that regard (restore or not).


STEP_LOCK_EVALUATION.md
aeyno, line 1: Does this file belong to the root of the repo ?

STEP_LOCK_EVALUATION.md
```diff
@@ -0,0 +1,103 @@
...
+- **Replication:** **N = 5 independent runs/arm** (distinct RNG). Fuzzing is high-variance; report
+  **median + [min,max]**, never a single run.
+- **Budget:** fixed wall-clock (24 h) and fixed cores/run, identical across arms; stagger to avoid
+  CPU contention.
+
+### 2.4 Metrics
+**Primary**
+- **Bad-switch reachability (legit):** hit-rate (k of N runs that crash within budget); **TTF**
+  (time-to-first) and **ETF** (execs-to-first).
+- **Corpus broken-rate:** `fuzz_campaigns/measure_hitchhiker.sh <dir> <bin> <sample>` — replays a
```
aeyno: What is `fuzz_campaigns/measure_hitchhiker.sh` ?

puffin/src/fuzzer/utils.rs
```diff
@@ -1,9 +1,34 @@
...
+    /// Per-stage anchor step lock. When set to `Some(k)`, every anchor drawn by
+    /// [`reservoir_sample`] -- and therefore all `choose*` helpers built on it -- is restricted to
+    /// step `k`. Set by [`crate::fuzzer::stages::StepLockedStackMutator`] so that all mutations
+    /// stacked in one mutational stage edit the SAME step. Global/Step *fan-out* (via
+    /// `find_all_term_filtered` / `find_all_sub_term_filtered`) does NOT consult this lock, so a
+    /// Global mutation still spreads its effect across the whole trace -- only its anchor choice is
+    /// confined to the locked step. Safe as a thread-local: LibAFL runs a single mutation thread
+    /// per fuzzer process (one process per core), so there is no cross-trace interference.
+    static STEP_LOCK: Cell<Option<StepIndex>> = const { Cell::new(None) };
+}
```
aeyno, line 19: Why not passing the step index as an argument of the mutations/resservoir sampling instead of using a static variable ?

aeyno (commented):
See comments

puffin/src/fuzzer/utils.rs
LCBH: We could instead create and use a metadata of the input but this would impact more the existing codebase. We  should go for it if you see any problem with the current solution though.

STEP_LOCK_EVALUATION.md
LCBH: Removed

puffin/src/fuzzer/utils.rs
aeyno: I'm good with it, I understand that passing some value through libAFL internal mechanism is not easy...

PR #535, opened 2026-09-11
[puffin:mutations] Set term-size caps to intended values

LCBH:
## Summary
Third PR in the mutation stack. #533 and #534 deliberately kept the term-size caps at the **conservative #531 values** so those PRs could be evaluated without the cap changes acting as a confounder. This PR sets the caps to their **intended values**, chosen above the current seed maxima across all three protocols (measured: TLS 324 nodes / 13 steps, OPC UA 137 / 10, SSH 285 / 10) to leave **headroom for future growth of the mappers** without another cap bump:

| cap | scope | old (#531) | new | rule / rationale |
|-----|-------|-----------|-----|------------------|
| `max_term_size` | a selected sub-term | 300 | **800** | `>= 2 * max_seed_term` (648); headroom for mapper growth |
| `max_result_term_size` | whole step recipe after replacement | 500 | **1000** | kept `>= max_term_size` so a max-size replacement sub-term fits |
| `max_result_trace_length` | max #steps | 15 | **20** | `>= max_seed_steps + 2` (15) |

The two size caps act on different scopes (a single sub-term vs the whole step recipe); beyond the invariants above they are a soft fuzzing-efficiency knob, to be measured in isolation. No logic change — only the three default values and their doc comments (the maintenance-rules table now also lists SSH).

## Base
Stacked on #534 (`pr/mutations/nb-per-stage-steplock`).

## Tests
`cargo test -p puffin --lib fuzzer::`, the tlspuffin seed-size tests, and the CVE reconstruction test all pass.


aeyno (approved):
LGTM

